PR #6, opened 2020-10-29
WIP: dependencies.yaml example

dprince:
Tested. Doesn't seem to work any different from the inline
requirements in the CSV.


dprince:
This doesn't work yet. Just an example of how we might wire in dependencies.yaml if we decide to go that route

dprince:
@abays Going to update this to try with   - type: olm.gvk

PR #7, opened 2020-10-30
Scripts for building and pushing images

abays:
Adds a script to build and push:

1. Operator image
2. Bundle image
3. Index image

Also has scripts to deploy and clean operator deployment via OLM.


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: quay.io/openstack-k8s-operators/osp-director-operator
+  newTag: 0.0.1
```
stuggi, line 8: in scripts/build_and_push_images.sh the default version is latest shouldn't this be in sync?

scripts/clean-olm.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -x
+
+TARGET_NAMESPACE=${TARGET_NAMESPACE:-"default"}
+CSV_VERSION=${CSV_VERSION:-"0.0.1"}
+
+oc delete -n ${TARGET_NAMESPACE} csv osp-director-operator.v${CSV_VERSION}
+oc delete -n ${TARGET_NAMESPACE} subscription osp-director-operator-subscription
+oc delete -n ${TARGET_NAMESPACE} catalogsource osp-director-operator-index
+oc delete -n ${TARGET_NAMESPACE} operatorgroup osp-director-operator-group
```
stuggi, line 10: isn't scripts/deploy-with-olm.sh and scripts/clean-olm.sh more a osp-director-dev-tools thing?

scripts/deploy-with-olm.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
stuggi, line 1: isn't scripts/deploy-with-olm.sh and scripts/clean-olm.sh more a osp-director-dev-tools thing?

config/manager/kustomization.yaml
abays: I haven't actually used this script yet to build and push to the `quay.io/openstack-k8s-operators` (just against my own registry).  Once we do that, this will be properly populated.

scripts/clean-olm.sh
abays: That's a fair point, as it certainly could be.  I was just following the pattern we used for the `openstack-cluster-operator`, where we had the build/deploy scripts in the operator repo itself: https://github.com/openstack-k8s-operators/openstack-cluster-operator/tree/master/scripts.  Since we're not merging CSVs, however, perhaps it makes sense to just move this script to the tools repo in this project's context.

dprince: I could go either way. I think that script could be useful as a standalone mechanism to install the eventual (working) operator outside of our dev-tooling environment too. So having it in the primary operator project also makes some sense.

scripts/deploy-with-olm.sh
dprince: @stuggi perhaps it is. Going to merge this as these are useful today. We can maybe move them later

PR #16, opened 2020-11-17
Controllers integration

abays:
A rough introduction of BaremetalSet and ProvisionServer controllers to our existing controllers


abays:
The go.mod and go.sum might need touching up before this properly compiles

abays:
We'll also get the Dockerfile working as @stuggi had it once we get everything integrated via this patch

controllers/provisionserver_controller.go
```diff
@@ -0,0 +1,310 @@
...
+)
+
+// ProvisionServerReconciler reconciles a ProvisionServer object
+type ProvisionServerReconciler struct {
+	client.Client
+	Kclient kubernetes.Interface
+	Log     logr.Logger
+	Scheme  *runtime.Scheme
+}
+
```
stuggi, line 51: same as above, I think we should have in all controllers[2]
GetClient() client.Client
GetKClient() kubernetes.Interface
GetLogger() logr.Logger
GetScheme() *runtime.Scheme

pkg/common/labels.go
```diff
@@ -24,3 +24,10 @@ func GetLabels(name string, appLabel string) map[string]string {
 		"app":   appLabel,
 	}
 }
+
+// InitLabelMap - Inititialise a label map to an empty map if it is nil.
+func InitLabelMap(m *map[string]string) {
+	if *m == nil {
+		*m = make(map[string]string)
+	}
+}
```
stuggi, line 33: Since it can be used more generic I creatd an InitMap [1], so we won't need this one. Also seems not to be used.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/map.go#L20

controllers/provisionserver_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	op, err := r.httpdConfigMapCreateOrUpdate(instance)
+
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if op != controllerutil.OperationResultNone {
+		r.Log.Info(fmt.Sprintf("Httpd ConfigMap %s successfully reconciled - operation: %s", instance.Name, string(op)))
+	}
+
```
stuggi, line 95: I think we could do all the configmap creation from the templates using template like https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/controllervm_controller.go#L124-L147 and use common.EnsureConfigMaps https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/configmap.go#L107 . but we can test/do that in a follow up. Note: when I did this for secrets like the controllervm use it, I added the same for configmaps, but it is not yet used and tested, but its how I did it for the e.g. nova-operator

controllers/provisionserver_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	// TODO: Myabe use filtering functions here since some resource permissions
+	// are now cluster-scoped?
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&ospdirectorv1beta1.ProvisionServer{}).
+		Owns(&appsv1.Deployment{}).
+		Owns(&corev1.ConfigMap{}).
+		Complete(r)
+}
+
+func (r *ProvisionServerReconciler) httpdConfigMapCreateOrUpdate(instance *ospdirectorv1beta1.ProvisionServer) (controllerutil.OperationResult, error) {
```
stuggi, line 143: see above configmap creation from template comment. this would remove the req for this

pkg/provisionserver/configmap.go
```diff
@@ -0,0 +1,34 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type provisionServerConfigOptions struct {
+	Port int
+}
+
+// HttpdConfigMap - custom httpd config map
+func HttpdConfigMap(cr *ospdirectorv1.ProvisionServer) *corev1.ConfigMap {
```
stuggi, line 15: see above configmap creation from template comment. this would remove the req for this

controllers/provisionserver_controller.go
```diff
@@ -0,0 +1,310 @@
...
+		replicas := int32(1)
+
+		deployment.Spec.Replicas = &replicas
+		deployment.Spec.Template.Spec = corev1.PodSpec{
+			HostNetwork: true,
+			Volumes:     volumes,
+			Containers: []corev1.Container{
+				{
+					Name:         "osp-httpd",
+					Image:        "quay.io/abays/httpd:2.4-alpine",
```
stuggi, line 195: should this be a CRD parameter?

controllers/provisionserver_controller.go
```diff
@@ -0,0 +1,310 @@
...
+					Env:          []corev1.EnvVar{},
+					VolumeMounts: volumeMounts,
+				},
+			},
+		}
+
+		initContainerDetails := provisionserver.InitContainer{
+			ContainerImage: "quay.io/abays/downloader:latest",
```
stuggi, line 203: same, should this be a CRD param?

pkg/util/env.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright 2020 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
```
stuggi, line 5: this is a dupe of https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/env.go

pkg/util/template_util.go
```diff
@@ -0,0 +1,104 @@
+package util
+
+import (
+	"bytes"
+	"fmt"
```
stuggi, line 5: this is a dupe of https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/template_util.go

templates/httpd/config/httpd.conf
```diff
@@ -0,0 +1,550 @@
+#
```
stuggi, line 1: for use the generic template functions it expects the files in templates/cr.Kind/config

config/samples/osp-director_v1beta1_provisionserver.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: osp-director.openstack.org/v1beta1
+kind: ProvisionServer
+metadata:
+  name: provisionserver-sample
```
stuggi, line 4: namespace: openstack ?

config/samples/osp-director_v1beta1_baremetalset.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: osp-director.openstack.org/v1beta1
+kind: BaremetalSet
+metadata:
+  name: baremetalset-sample
```
stuggi, line 4: namespace: openstack ?

controllers/provisionserver_controller.go
abays: Sounds good to me.  This was just a quick approach along the "fail fast" track to just see if all of this would even work.

Dockerfile
```diff
@@ -26,41 +12,29 @@ RUN go mod download
...
     OPERATOR_TEMPLATES=/usr/share/osp-director-operator/templates/ \
-    WATCH_NAMESPACE=openstack
-
-# install our bindata
-RUN  mkdir -p ${OPERATOR_BINDATA_DIR}
-COPY --from=builder ${DEST_ROOT}/bindata ${OPERATOR_BINDATA_DIR}
-
-RUN  mkdir -p ${OPERATOR_TEMPLATES}
-COPY --from=builder ${DEST_ROOT}/templates ${OPERATOR_TEMPLATES}
+    WATCH_NAMESPACE=openstack,openshift-machine-api
```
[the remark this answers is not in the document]
abays: I added a `cluster_role_binding.yaml` (and associated `kustomization.yaml` line) that might not be necessary, but the `kubebuilder:rbac` annotation changes were required to get past permission errors in the controllers.

controllers/provisionserver_controller.go
abays: There's a design discussion to be had around this Apache container and the `downloader` init container image.  The current approach emulates what Metal3 did for their provisioning Apache server.  The `downloader` expects a `.gz` image and creates a compressed `qcow2` image at `/usr/local/apache/htdocs/images/<image name>/compressed-<image name>.qcow2`.  This `/usr/local/apache/htdocs` is then the hard-coded `httpd.conf` default for `DocumentRoot`, which the `httpd:2.4-alpine` container uses to serve the image.  Thus the paths are predictable and can be surfaced in the other operator logic.

We could, of course, make a lot (if not all) of this customizable via the CRD.  It's ugly to be using my quay images right now, but eventually these would come from some official registry in their final form.  I guess I assumed the user would care which is ultimately used, but not be concerned with how its stored or served (aside from choosing the port).  That being said, there may be things I am not considering in an OSP context, as I have simply approached this a variation of the Metal3 OCP implementation for now.

PR #27, opened 2020-11-19
Wait for prov server pod = running

abays:
Metal3 will attempt to validate the image URL passed to it via BareMetalHost, so we need to wait until the ProvisionServer pod is running so that the image is actually available.

### DIFF
--- a/controllers/provisionserver_controller.go
+++ b/controllers/provisionserver_controller.go
@@ -285,6 +285,12 @@ func (r *ProvisionServerReconciler) getProvisionServerProvisioningIP(instance *o
 	pod := podList.Items[0]
 	r.Log.Info(fmt.Sprintf("Found pod %s on node %s", pod.Name, pod.Spec.NodeName))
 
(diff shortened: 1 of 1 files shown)


controllers/provisionserver_controller.go
```diff
@@ -285,6 +285,12 @@ func (r *ProvisionServerReconciler) getProvisionServerProvisioningIP(instance *o
 	pod := podList.Items[0]
 	r.Log.Info(fmt.Sprintf("Found pod %s on node %s", pod.Name, pod.Spec.NodeName))
 
+	// If the pod is not running yet, we don't actually have the server available
+	// on the provisioning IP, so just stop here for now
+	if pod.Status.Phase != corev1.PodRunning {
+		return "", nil
```
dprince, line 291: Do you want to Requeue here instead?

So this perhaps:
return ctrl.Result{RequeueAfter: time.Second * 30}, err

PR #31, opened 2020-11-23
Connect OSP controller VMs to OSP network

stuggi:
Use nmstate to configure the interface/bridge to the OSP network
using nodenetworkconfigurationpolicies. This interface is then used
in a network a network-attachment-definitions to connect the VM
to the OSP network. On top of this network we can use os-connect-config
to get the predictiable IP configured.


bindata/network/01-network-configurtion-policy.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: nmstate.io/v1alpha1
+kind: NodeNetworkConfigurationPolicy
+metadata:
+  name: {{ .BridgeName }}
+spec:
+  nodeSelector:
+    node-role.kubernetes.io/worker: ""
+  desiredState:
+    interfaces:
```
dprince, line 9: So instead of prescribing linux bridge I wonder if it might be better to just pass in the entire desiredState/interfaces section via our CRD. Essentially just directly exposing the nmstate CRD's interface via ours? I'm not against using a bridge for now as our default, perhaps we could just have that in our sample config?

stuggi: added it, please check latest push.

controllers/controllervm_controller.go
```diff
@@ -267,15 +271,12 @@ func (r *ControllerVMReconciler) networkCreateAttachmentDefinition(instance *osp
...
-			r.Log.Error(err, "Failed to render network manifests : %v")
-			return err
-		}
-		objs = append(objs, manifests...)
+	manifests, err := bindatautil.RenderDir(filepath.Join(ManifestPath, "network"), data)
+	if err != nil {
+		r.Log.Error(err, "Failed to render network manifests : %v")
+		return err
 	}
+	objs = append(objs, manifests...)
```
dprince, line 279: Just a question. Why did you move the append for objs here and comment them out in the loop below?

stuggi: the objs here is the same as before. for now we don't have an array of networks in the CRD, so no loop. The ops bellow were just not required as we don't need to add the modified obj again to objs. I missed to delete to commented like completely

dprince (approved):

PR #33, opened 2020-11-24
BMSet bootstrapping

abays:
Working version of BaremetalSet bootstrapping + basic scaling


api/v1beta1/baremetalset_types.go
```diff
@@ -26,10 +26,23 @@ type BaremetalSetSpec struct {
...
 
 // BaremetalSetStatus defines the observed state of BaremetalSet
 type BaremetalSetStatus struct {
+	BaremetalHosts map[string]BaremetalHostStatus `json:"baremetalHosts"`
+}
+
+// BaremetalHost represents a BMH resource allocated by the BaremetalSet
+type BaremetalHostStatus struct {
+	SecretName string `json:"secretName"`
+	MgmtIP     string `json:"mgmtIP"`
```
stuggi, line 45: should we add the metal3v1alpha1.BareMetalHost power status here?

controllers/baremetalset_controller.go
```diff
@@ -90,76 +102,139 @@ func (r *BaremetalSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		if len(result) > 0 {
+			return result
+		}
+		return nil
+	})
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ospdirectorv1beta1.BaremetalSet{}).
 		Owns(&ospdirectorv1beta1.ProvisionServer{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}},
```
stuggi, line 234: what happens when the bmh secret changes? right now I don't see a need to reconcile when a bmh secret changes. We can not update e.g. ssh auth_keys via this controller to access the server. But maybe I miss something

But we might want to reconcile on our metal3v1alpha1.BareMetalHost object and as mentioned above add the power status in our the cr status?

controllers/baremetalset_controller.go
```diff
@@ -90,76 +102,139 @@ func (r *BaremetalSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		return ctrl.Result{}, err
 	}
 
 	return ctrl.Result{}, nil
 }
 
 // SetupWithManager - prepare controller for use with operator manager
 func (r *BaremetalSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	openshiftMachineApiSecretsFn := handler.ToRequestsFunc(func(o handler.MapObject) []reconcile.Request {
+		cc := mgr.GetClient()
```
stuggi, line 204: we don't have to get the client r.Client can be used

controllers/baremetalset_controller.go
```diff
@@ -90,76 +102,139 @@ func (r *BaremetalSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		result := []reconcile.Request{}
+		secret := &corev1.Secret{}
+		key := client.ObjectKey{Namespace: o.Meta.GetNamespace(), Name: o.Meta.GetName()}
+		err := cc.Get(context.Background(), key, secret)
+		if err != nil && !errors.IsNotFound(err) {
+			r.Log.Error(err, "Unable to retrieve Secret %v")
+			return nil
+		}
+
+		label := secret.ObjectMeta.GetLabels()
```
stuggi, line 214: as we reconcile in secret objects, all this could be just 
`label := o.Meta.GetLabels()`

stuggi (commented):
need to have another look in more detail, but added some comments/thoughts

controllers/baremetalset_controller.go
```diff
@@ -187,3 +262,270 @@ func (r *BaremetalSetReconciler) provisionServerCreateOrUpdate(instance *ospdire
...
+		// We're done
+		r.Log.Info("All requested BaremetalHost replicas have been provisioned")
+		return nil
+	}
+
+	// We have new replicas requested, so search for baremetalhosts that don't have consumerRef or Online set
+	availableBaremetalHosts := []string{}
+	baremetalHostsList := &metal3v1alpha1.BareMetalHostList{}
+	listOpts := []client.ListOption{
+		client.InNamespace("openshift-machine-api"),
```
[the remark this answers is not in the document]
abays: Yep, we can definitely do that as we flesh this out.

api/v1beta1/baremetalset_types.go
abays: Seems reasonable to me :+1:

controllers/baremetalset_controller.go
abays: This watch might be the wrong approach indeed.  A result of my novice operator knowledge that will hopefully improve. :-)

But yes, watching the `BareMetalHost` object seems proper to add before we finish this draft.

controllers/baremetalset_controller.go
abays: Ack

controllers/baremetalset_controller.go
abays: Ack

templates/baremetalset/config/userData
```diff
@@ -0,0 +1,26 @@
...
+  - path: /etc/sysconfig/network-scripts/ifcfg-{{ .MgmtInterface }}
+    content: |
+      BOOTPROTO=none
+      DEVICE={{ .MgmtInterface }}
+      IPADDR={{ .MgmtIp }}
+      PREFIX={{ .MgmtNetmask }}
+      ONBOOT=yes
+      STARTMODE=auto
+      TYPE=Ethernet
+      USERCTL=no
```
stuggi, line 26: any reason to use those cmds and write the nw config file instead of using networkData for nw config?

abays: This is just what I was able to get working.  I had tried following this to no avail:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_cloud-init_for_rhel_8/configuring-cloud-init_cloud-content#setting-up-a-static-networking-configuration-with-cloud-init_configuring-cloud-init

If there is a better option, such as `networkData`, let's do that instead.

controllers/baremetalset_controller.go
dprince: also, I think there is a performance hit if you watch objects across namespaces.

controllers/baremetalset_controller.go
```diff
@@ -19,8 +19,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"net"
+	"strings"
 	"time"
 
+	"github.com/apparentlymart/go-cidr/cidr"
```
dprince, line 26: I hadn't been using this for my IP/CIDR code. But it tidies things up and it looks like some other dependencies use it as well so lets add it in

controllers/baremetalset_controller.go
```diff
@@ -187,3 +254,322 @@ func (r *BaremetalSetReconciler) provisionServerCreateOrUpdate(instance *ospdire
...
+		}
+	}
+
+	_, network, err := net.ParseCIDR(instance.Spec.MgmtCIDR)
+
+	if err != nil {
+		return "", err
+	}
+
+	for i := uint64(2); i <= cidr.AddressCount(network); i++ {
```
dprince, line 553: I think we want this to be 'i < cidr.AddressCount(network);' as the last IP say 256 in a /24 is reserved

abays: Yes, good catch.  The last one is usually the broadcast IP.

controllers/baremetalset_controller.go
abays: There indeed may be a performance hit.  I am just replicating the pattern we used for the compute node operator.  This is also what the experts suggested when I asked them about it in one of the operator development meetings in the past few weeks.

PR #39, opened 2020-12-03
Baremetal cloud-init via networkData

abays:
Uses Metal3's `NetworkData` BMH field to set networking config on the hosts (instead of hacking `UserData`).  IP-generation code is now just for demo purposes, as we will be handling this externally soon enough.


controllers/baremetalset_controller.go
```diff
@@ -535,27 +573,40 @@ func (r *BaremetalSetReconciler) deleteOwnerRefLabeledObjects(instance *ospdirec
...
+func findIPs(instance *ospdirectorv1beta1.BaremetalSet, bmhName string) (ip string, gateway string, netmask string, err error) {
+	_, network, err := net.ParseCIDR(instance.Spec.MgmtCIDR)
+
+	if err != nil {
+		return "", "", "", err
 	}
 
-	_, network, err := net.ParseCIDR(instance.Spec.MgmtCIDR)
+	// Gateway will just be first address in the network for now
+	gw, err := cidr.Host(network, 1)
```
dprince, line 585: we may need to revisit this as I'm not sure Management networks would be set as a default gateway. If the network is just used for accessing machines they may not need a default GW/route on that network. For now as this is the only way to get to the machine we can use it though.

templates/baremetalset/cloudinit/networkdata
```diff
@@ -0,0 +1,11 @@
...
+- name: {{ .MgmtInterface }}
+  id: {{ .MgmtInterface }}
+  type: vif
+networks:
+- netmask: {{ .MgmtNetmask }}
+  link: {{ .MgmtInterface }}
+  id: {{ .MgmtInterface }}
+  ip_address: {{ .MgmtIp }}
+  type: ipv4
+  gateway: {{ .MgmtGateway }}
```
dprince, line 11: ditto here on gateway. we may need to make it configurable and possible unset by default for the management network.

dprince (approved):

controllers/baremetalset_controller.go
```diff
@@ -535,27 +573,40 @@ func (r *BaremetalSetReconciler) deleteOwnerRefLabeledObjects(instance *ospdirec
...
 
-// Get the IP from the management network for a particular BaremetalHost in the BaremetalSet
-func findNewMgmtIP(instance *ospdirectorv1beta1.BaremetalSet, bmhName string) (string, error) {
-	// First check if we already have an IP for this host
-	if bmhStatus, ok := instance.Status.BaremetalHosts[bmhName]; ok {
-		if bmhStatus.MgmtIP != "" {
-			return bmhStatus.MgmtIP, nil
-		}
+// Get the IP, gateway and netmask from the management network for a particular BaremetalHost in the BaremetalSet
+func findIPs(instance *ospdirectorv1beta1.BaremetalSet, bmhName string) (ip string, gateway string, netmask string, err error) {
```
stuggi, line 577: maybe we return an "ipset" instead of 3 separate, but as this will change with the discussed network CRDs its ok

controllers/baremetalset_controller.go
```diff
@@ -430,7 +454,7 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
 	}
 
 	// Set status (add this BaremetalHost entry)
-	r.setBaremetalHostStatus(instance, foundBaremetalHost, secretName, mgmtIP)
+	r.setBaremetalHostStatus(instance, foundBaremetalHost, userDataSecretName, networkDataSecretName, mgmtIP)
```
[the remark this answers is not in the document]
abays: Good idea

controllers/baremetalset_controller.go
```diff
@@ -366,22 +360,48 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
...
+	// User data cloud-init secret
 	templateParameters := make(map[string]string)
 	templateParameters["AuthorizedKeys"] = strings.TrimSuffix(string(sshSecret.Data["authorized_keys"]), "\n")
+
+	userDataSecretName := fmt.Sprintf(baremetalset.CloudInitUserDataSecretName, instance.Name, bmh)
+
+	st := common.Template{
+		Name:           userDataSecretName,
+		Namespace:      "openshift-machine-api",
+		Type:           common.TemplateTypeConfig,
```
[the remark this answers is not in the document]
abays: It seems to work, but I'd like to do it the proper way as per your common Secret logic.  I will use the `TemplateTypeNone` once it is available

PR #45, opened 2020-12-10
RHEL image downloader container

abays:
First draft of init container that downloads images from compute node provisioning


stuggi (approved):
lgtm

PR #46, opened 2020-12-10
BaremetalSet BMH selection

abays:


api/v1beta1/baremetalset_types.go
```diff
@@ -47,6 +52,70 @@ type BaremetalHostStatus struct {
...
+type HardwareReqs struct {
+	CPUReqs  CPUReqs  `json:"cpuReqs,omitempty"`
+	MemReqs  MemReqs  `json:"memReqs,omitempty"`
+	DiskReqs DiskReqs `json:"diskReqs,omitempty"`
+}
+
+// CPUReqs defines specific CPU hardware requests
+type CPUReqs struct {
+	// Arch is a scalar (string) because it wouldn't make sense to give it an "exact-match" option
+	Arch     string      `json:"arch,omitempty"`
```
stuggi, line 65: afaic right now only x86_64 and ppc64le computes are supported with OSP, should't we limit it to those archs?

https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.1/html-single/director_installation_and_usage/index#compute-node-requirements

api/v1beta1/baremetalset_types.go
```diff
@@ -47,6 +52,70 @@ type BaremetalHostStatus struct {
...
+	CountReq CPUCountReq `json:"countReq,omitempty"`
+	MhzReq   CPUMhzReq   `json:"mhzReq,omitempty"`
```
stuggi, line 67: does the hw status reflect numa nodes? that'd be some good selection

stuggi (commented):
in general I can see the following node selection req:
- any of the free bmh nodes
- any of the free bmh nodes which has some specific hw like in this change for cpu/mem/disk + sriov, gpu, ...
- any bmh of a group of bmh nodes by name, e.g. worker-1..10 are openstack nodes
- strict order of a bmh of a group of bmh nodes by name, like worker-1 ..10 are openstack nodes, if Replicas: 5, it should start with 1..5

api/v1beta1/baremetalset_types.go
abays: If that's indeed the case, that sounds reasonable to me

api/v1beta1/baremetalset_types.go
abays: Unfortunately numa information is not gathered by the IPA, or at least is not placed in the BMH status

abays:
> in general I can see the following node selection req:
> 
>     * any of the free bmh nodes
> 
>     * any of the free bmh nodes which has some specific hw like in this change for cpu/mem/disk + sriov, gpu, ...
> 
>     * any bmh of a group of bmh nodes by name, e.g. worker-1..10 are openstack nodes
> 
>     * strict order of a bmh of a group of bmh nodes by name, like worker-1 ..10 are openstack nodes, if Replicas: 5, it should start with 1..5

We already had item 1, and we now have item 2 to the extent that we can with the information expose in Metal3 BaremetalHost resources.  Item 3 is now possible via labeling BaremetalHosts arbitrarily, and then including that/those label(s) in the `bmhLabelSelector` map in the CRD.  Item 4 was addressed by enforcing a strict ordering of available BaremetalHost resources, sorted by name.

dprince:
The type structs here are nice in that they will allow us to validate some fields. An alternate implementation might be a single string used to generically match any field (perhaps the implementation marshalls to json or yaml first). That would sort of future proof the effort but it also doesn't let us validate any of the user input and is perhaps harder to use (would require more up front examples).

dprince (approved):

stuggi (approved):

PR #47, opened 2020-12-11
install repositories in bms cloudinit userdata

stuggi:


templates/baremetalset/cloudinit/userdata
```diff
@@ -7,4 +7,7 @@ users:
     ssh-authorized-keys: {{ .AuthorizedKeys }}
     sudo: ['ALL=(ALL) NOPASSWD:ALL']
     shell: /bin/bash
+runcmd:
+  - dnf install -y http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
```
dprince, line 11: @stuggi private URL here. Should we make a parameter for this instead?

stuggi: yes, in the end. this is just to test/make initial deploy workflow. in the end we need to add support for rhsm/satellite based registration.

abays:
lgtm

PR #53, opened 2020-12-16
Improved provisioning IP discovery

abays:


abays:
Closing in favor of the approach in https://github.com/openstack-k8s-operators/osp-director-operator/pull/60

PR #57, opened 2020-12-17
Integrate BaremetalSet with OvercloudIpset

dprince:
Still testing...


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/osp-director-operator
-  newTag: 0.0.1
+  newName: quay.io/dprince/osp-director-operator
+  newTag: 0.0.9
```
abays, line 8: Should probably not merge these changes.  We have this in the `.gitignore`, but somehow it isn't being honored.

dprince: Fixed in latest comment. Thanks

controllers/baremetalset_controller.go
```diff
@@ -440,13 +473,16 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
 	}
 
 	sts = append(sts, userDataSt)
+	ip, network, _ := net.ParseCIDR(ipset.Status.HostIPs[bmhName].IPAddresses["ctlplane"]) // We use ctlplane as the MgmtNetwork too for now
+
+	r.Log.Info(fmt.Sprintf("Network: %s", network.String()))
```
abays, line 478: Is this left here intentionally?  If so, might be good to include BaremetalSet name and `bmhName` to identify the association with the network.

controllers/baremetalset_controller.go
```diff
@@ -440,13 +473,16 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
...
 
 	// Network data cloud-init secret
 	templateParameters = make(map[string]string)
-	templateParameters["MgmtIp"] = mgmtIP
+	templateParameters["MgmtIp"] = ip.String()
 	templateParameters["MgmtInterface"] = instance.Spec.MgmtInterface
-	templateParameters["MgmtGateway"] = gwIP
-	templateParameters["MgmtNetmask"] = netmask
+	templateParameters["MgmtGateway"] = "192.168.1.1" //FIXME
+	templateParameters["MgmtNetmask"] = fmt.Sprintf("/%s", strings.Split(network.String(), "/")[1])
```
abays, line 485: This format might not work properly with RHEL's cloud-init, unless you've tried it and found it to be okay.  I say that only because the examples I found all used `X.X.X.X` notation for IPv4 (never found an IPv6 example, however).

controllers/overcloudipset_controller.go
```diff
@@ -85,48 +86,58 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
-		_, cidr, _ := net.ParseCIDR(network.Spec.Cidr)
-		if network.Name == "ctlplane" {
-			ctlplaneCidr = network.Spec.Cidr
-		}
-		start := net.ParseIP(network.Spec.AllocationStart)
-		end := net.ParseIP(network.Spec.AllocationEnd)
+			network := &ospdirectorv1beta1.OvercloudNet{}
+			err := r.Client.Get(context.TODO(), types.NamespacedName{Name: netName, Namespace: instance.Namespace}, network)
+			if err != nil {
+				if k8s_errors.IsNotFound(err) {
```
abays, line 110: Should we log something here to inform the user that the required network is missing?  I admit, though, that I'm not sure who will be responsible for creating the networks in the first place.  Is it the user, or perhaps another controller?  Dev-tools for now?

controllers/baremetalset_controller.go
dprince: let me research this again. I meant to do this as a separate patch actually and it slipped in on a rebase. Probably best to just leave it as a dotted quad format

controllers/overcloudipset_controller.go
dprince: ack. lets log it.

controllers/baremetalset_controller.go
dprince: Latest patch has this resolved.

controllers/overcloudipset_controller.go
dprince: got it

PR #63, opened 2020-12-18
Fix controller networkdata secret generation

stuggi:
The networkdata rendering is running for all controllers with the
same ControllerIP template parameter which results in with multiple
controllers that all have the same ip on the osp network. Now we
make sure that the secret gets rendered with uniq ips.

### DIFF
--- a/controllers/controllervm_controller.go
+++ b/controllers/controllervm_controller.go
@@ -122,7 +122,7 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	templateParameters := make(map[string]string)
 	templateParameters["AuthorizedKeys"] = strings.TrimSuffix(string(sshSecret.Data["authorized_keys"]), "\n")
 
-	sts := []common.Template{
+	cloudinit := []common.Template{
 		// CloudInitSecret
 		{
 			Name:           fmt.Sprintf("%s-cloudinit", instance.Name),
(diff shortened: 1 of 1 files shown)


controllers/controllervm_controller.go
```diff
@@ -223,6 +227,7 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	if instance.Status.BaseImageDVReady {
 		// Generate the Contoller networkdata secrets
 		for _, ctl := range controllerDetails {
+			r.Log.Info(fmt.Sprintf("DEBUG: %v", ctl))
```
dprince, line 230: do we want to leave this in?

controllers/controllervm_controller.go
```diff
@@ -152,34 +157,33 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+			DomainName:        hostname,
+			DomainNameUniq:    fmt.Sprintf("%s-%s", hostname, instance.UID[0:4]),
+			IPAdress:          fmt.Sprintf("192.168.25.1%d", i),
+			NetworkDataSecret: networkDataSecretName,
+		}
 
-	err = common.EnsureSecrets(r, instance, sts, &envVars)
-	if err != nil {
-		return ctrl.Result{}, nil
+		templateParameters["ControllerIP"] = controllerDetails[hostname].IPAdress
```
dprince, line 170: I had noticed this last week. No need to fix it here but I think we should rename this IPAddress.

dprince (approved):

abays:
This looks good to me

controllers/controllervm_controller.go
stuggi: no, removing it

controllers/controllervm_controller.go
stuggi: yes sounds good. we can do it with the IPAM integration

PR #70, opened 2020-12-22
Specific BMH scale-down

abays:
Allows targeted BaremetalHost scale-down (via annotation) for BaremetalSets

### DIFF
--- a/controllers/baremetalset_controller.go
+++ b/controllers/baremetalset_controller.go
@@ -299,43 +299,42 @@ func (r *BaremetalSetReconciler) ensureBaremetalHosts(instance *ospdirectorv1bet
 	// How many new BaremetalHost de-allocations do we need (if any)?
 	oldBmhsToRemoveCount := len(existingBaremetalHosts) - instance.Spec.Replicas
 
-	for i := 0; i < oldBmhsToRemoveCount; i++ {
-		// First choose BaremetalHosts to remove from the prepared list of BaremetalHosts
-		// that have the "osp-director.openstack.org/baremetalset-delete-baremetalhost=yes" annotation
+	if oldBmhsToRemoveCount > 0 {
+		oldBmhsRemovedCount := 0
 
-		if len(removalAnnotatedBaremetalHosts) > 0 {
-			err := r.baremetalHostDeprovision(instance, removalAnnotatedBaremetalHosts[0])
+		for i := 0; i < oldBmhsToRemoveCount; i++ {
+			// First choose BaremetalHosts to remove from the prepared list of BaremetalHosts
+			// that have the "osp-director.openstack.org/baremetalset-delete-baremetalhost=yes" annotation
 
-			if err != nil {
-				return err
-			}
+			if len(removalAnnotatedBaremetalHosts) > 0 {
+				err := r.baremetalHostDeprovision(instance, removalAnnotatedBaremetalHosts[0])
 
-			// Remove the removal-annotated BaremetalHost from the existingBaremetalHosts map
-			delete(existingBaremetalHosts, removalAnnotatedBaremetalHosts[0])
+				if err != nil {
+					return err
+				}
 
-			// Remove the removal-annotated BaremetalHost from the removalAnnotatedBaremetalHosts list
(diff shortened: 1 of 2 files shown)


controllers/baremetalset_controller.go
```diff
@@ -274,32 +274,67 @@ func (r *BaremetalSetReconciler) ensureBaremetalHosts(instance *ospdirectorv1bet
...
 	// Find the current BaremetalHosts belonging to this BaremetalSet
 	for _, baremetalHost := range baremetalHostsList.Items {
 		if baremetalHost.Spec.ConsumerRef != nil && (baremetalHost.Spec.ConsumerRef.Kind == instance.Kind && baremetalHost.Spec.ConsumerRef.Name == instance.Name && baremetalHost.Spec.ConsumerRef.Namespace == instance.Namespace) {
 			r.Log.Info(fmt.Sprintf("Existing BaremetalHost: %s", baremetalHost.ObjectMeta.Name))
-			existingBaremetalHosts = append(existingBaremetalHosts, baremetalHost.ObjectMeta.Name)
+			// Storing the existing BaremetalHosts in a map like this just makes selected replica scale-down easier below
+			existingBaremetalHosts[baremetalHost.ObjectMeta.Name] = baremetalHost.ObjectMeta.Name
+
+			// Prep for possible replica scale-down below
+			if val, ok := baremetalHost.Annotations[baremetalset.BaremetalHostRemovalAnnotation]; ok && (val == "yes" || val == "true") {
```
stuggi, line 288: we probably should use https://golang.org/pkg/strings/#EqualFold or strings.ToLower(val) to do case-insensitivity  compare, just to make sure someone did not set BaremetalHostRemovalAnnotation to Yes/True?

controllers/baremetalset_controller.go
```diff
@@ -274,32 +274,67 @@ func (r *BaremetalSetReconciler) ensureBaremetalHosts(instance *ospdirectorv1bet
...
-				existingBaremetalHosts = []string{}
+				removalAnnotatedBaremetalHosts = []string{}
+			}
+
+			// We removed a removal-annotated BaremetalHost, so continue
+			continue
+		}
+
+		// We did NOT find a removal-annotated BaremetalHost to remove above, so we pick the first one
+		// in the existingBaremetalHosts list
```
stuggi, line 328: We probably should not do any action if the host to remove was not annotated for removal to make sure we do not remove a node which is not expected to be deleted. The workflow today in tripleo also requires specify a specific node to delete.

controllers/baremetalset_controller.go
abays: Sure, that's a good point

controllers/baremetalset_controller.go
```diff
@@ -274,33 +274,67 @@ func (r *BaremetalSetReconciler) ensureBaremetalHosts(instance *ospdirectorv1bet
...
-	// Deallocate existing BaremetalHosts to match the requested replica count, if necessary.
-	// TODO: Add something to CR to allow user to specify which BaremetalHosts are removed
-	for i := len(existingBaremetalHosts); i > instance.Spec.Replicas; i-- {
-		for _, oldBmhName := range existingBaremetalHosts {
-			// Just delete the first one we find until the TODO from above is implemented
-			err := r.baremetalHostDeprovision(instance, oldBmhName)
+	// Deallocate existing BaremetalHosts to match the requested replica count, if necessary.  First we
+	// choose BaremetalHosts with the "osp-director.openstack.org/baremetalset-delete-baremetalhost=yes"
+	// annotation, then, if there are still BaremetalHosts left to deprovision, we choose randomly from the
+	// existingBaremetalHosts map (as maps are unordered in Golang, we don't know which one will be chosen)
```
stuggi, line 297: we should remove the choose randomly part of the comment.

stuggi (commented):
lgtm, just one small change to remove the "random" part from the comment

controllers/baremetalset_controller.go
abays: Good catch

stuggi (approved):
lgtm

dprince:
lgtm

PR #71, opened 2021-01-05
Allow uncompressed source prov images

abays:
Removes the requirement that ProvisionServer source qcow2s be compressed, allowing either compressed or uncompressed.


stuggi (approved):
lgtm

dprince (commented):
In the case of uncompressed there are still a few variable names (RHEL_IMAGE_FILENAME_COMPRESSED for example) below that mention compressed. I don't think that functionally matters but it is confusing. Should we add a comment or otherwise clarify those for future reference?

abays:
Yes, we can clarify it in the comments.  But we do indeed compress the image after we download it, and store it as such.

dprince:
lgtm, thanks for the updates

PR #74, opened 2021-01-07
IPAM BMSet scale functionality

abays:
Currently contains fixes for duplicate hostname/IP issues, as well as another problem involving a stuck reconciliation loop for BaremetalSets.  However, it does not clean-up the respective hostnames/IPs in OvercloudIpSet status nor OvercloudNet status during BaremetalSet scale-down.  We are leaving it this way for now.

### DIFF
--- a/controllers/overcloudipset_controller.go
+++ b/controllers/overcloudipset_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/go-logr/logr"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
@@ -100,10 +101,6 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 		// iterate over the requested Networks
 		for _, netName := range instance.Spec.Networks {
 
-			if instance.Status.HostIPs[hostname].IPAddresses[netName] != "" {
-				continue
-			}
-
 			network := &ospdirectorv1beta1.OvercloudNet{}
 			err := r.Client.Get(context.TODO(), types.NamespacedName{Name: netName, Namespace: instance.Namespace}, network)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


controllers/overcloudipset_controller.go
```diff
@@ -122,20 +122,35 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
-			// record the reservation on the OvercloudNet
-			network.Status.Reservations = reservation
-			err = r.Client.Status().Update(context.TODO(), network)
-			if err != nil {
-				r.Log.Error(err, "Failed to update OvercloudNet status %v")
-				return ctrl.Result{}, err
+			// Do we already have a reservation for this hostname on the network?
+			for _, reservation := range network.Status.Reservations {
+				if reservation.Hostname == hostname {
+					reservationIP = reservation.IP
```
stuggi, line 130: I think with this we want to remove the check in L103-L105 and return the already existing reservation here.
```
			if instance.Status.HostIPs[hostname].IPAddresses[netName] != "" {
				continue
			}
```
I have seen issues with our current version and the above check/continue as ctrlplanCidr might not get set and added to the environment file.

abays: Yes, I think that might be right.  If the `hostname` is already present in the OvercloudIPSet status, then we can assume the same reservation is also already stored in the OvercloudNet.  But should we return, or do something else (other than `continue`)?  Do we want the `EnsureConfigMaps` logic to be processed?  A return would skip that logic.

stuggi: atm I think we want it processed with the existing information

dprince: Agree that you can probably drop the check at line 103 with this new approach. It probably doesn't hurt to recreate the ConfigMaps

abays: Sounds good.  I will remove it and run the change through my tests.

PR #76, opened 2021-01-11
Use OvercloudIpset for controller_vm IPAM

dprince:
Also includes new implementation of hostnameGeneration.


pkg/common/hostname.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// HostnameStore -
+type HostnameStore interface {
```
dprince, line 26: @abays hey, I ninja'd in your function here and modified it to use this interface for hostnames. See the ControllerVM object above where I implemented the interface. This way both BaremetalSet and ControllerVM can share it but have different Status's to store the results.

controllers/controlplane_controller.go
```diff
@@ -128,7 +128,9 @@ func (r *ControlPlaneReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	// Create or update the controllerVM CR object
 	ospControllerVM := &ospdirectorv1beta1.ControllerVM{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-controllervm", instance.Name),
+			// FIXME: currently we use the name of the resource as the basename
+			Name: "controller",
```
dprince, line 132: @stuggi @abays in the BaremetalSet version of this patch (we merged last week) I had used the name of the resource as the basis of the hostname. So for baremetalSets 'compute', 'compute1', etc. For controllers we are creating just a single ctlplane right? And that creates the controlplane_vm's... so is it okay to just hard code this? Reason I ask is the alternate would be to use the name of the ControlPlane resource but we name it 'overcloud' so I'm not sure it fits...

stuggi: In my patch that I work on atm, I changed the baremetalset hostname to use the role name instead of the CR name,
which I think more alligns with the tripleo side. I planned to do the same for the controller, add a role param to the
CRD and use it for the hostname and set the HostnameFormat for the role. With this you could also use customized roles.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/osp-director-operator
-￼ newTag: 0.0.1
+  newName: quay.io/dprince/osp-director-operator
+  newTag: 0.0.15
```
stuggi, line 8: not include this change in the patch

config/rbac/role.yaml
```diff
@@ -307,6 +307,7 @@ rules:
   - overcloudipsets/status
   verbs:
   - get
+  - patc
```
stuggi, line 310: seems to be a typo?

controllers/controllervm_controller.go
```diff
@@ -83,6 +86,9 @@ func (r *ControllerVMReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,namespace=openstack,resources=network-attachment-definitions,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=kubevirt.io,namespace=openstack,resources=virtualmachines,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=nmstate.io,resources=nodenetworkconfigurationpolicies,verbs=create;delete;get;list;patch;update;watch
+// +kubebuilder:rbac:groups=osp-director.openstack.org,resources=overcloudipsets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=osp-director.openstack.org,resources=overcloudipsets/finalizers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=osp-director.openstack.org,resources=overcloudipsets/status,verbs=get;update;patc
```
stuggi, line 91: should be patc -> patch

controllers/controllervm_controller.go
```diff
@@ -156,18 +177,30 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		r.Log.Info(fmt.Sprintf("Controlplane VM hostname set to %s", hostname))
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		//hostname := fmt.Sprintf("controller-%d", i)
 		networkDataSecretName := fmt.Sprintf("%s-%s-networkdata", instance.Name, hostname)
 
+		ip, _, _ := net.ParseCIDR(ipset.Status.HostIPs[hostKey].IPAddresses["ctlplane"]) // We use ctlplane as the MgmtNetwork too for now
+		// FIXME where is the netmask set!?
```
stuggi, line 191: the controllers use cloud-init networking config v2 https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v2.html#common-properties-for-all-device-types , so it uses CIDR notation https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/controllervm/cloudinit/networkdata#L4

controllers/controllervm_controller.go
```diff
@@ -240,6 +273,39 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	overcloudIPSet := &ospdirectorv1beta1.OvercloudIPSet{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      instance.Name, // used as the base name of server instances
+			Namespace: instance.ObjectMeta.Namespace,
+		},
+	}
+
+	op, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, overcloudIPSet, func() error {
+		overcloudIPSet.Spec.Networks = []string{"ctlplane"} //FIXME provide way to expose more networks via CR
+		overcloudIPSet.Spec.Role = "controller"             // NOTE: will VM's always be controllers
```
stuggi, line 294: we should add the role to the CRD that we could support custom roles

controllers/controlplane_controller.go
dprince: @stuggi How were you planning on making hostname unique then? I had considered it, but as the 'Name' is guaranteed to be unique it seemed to fit instead.

dprince: Looks like kubebuilder has validations that support unique. I've not tried them yet however...

stuggi: @dprince same like we have atm, rolename + index, just replacing instance.Name with the role name.

pkg/common/hostname.go
abays: Cool!  This abstraction is a good idea

dprince:
@abays lgtm on this

config/manager/kustomization.yaml
dprince: Roger that. Removed in latest patch.

controllers/controllervm_controller.go
dprince: Got it in the latest push.

controllers/controllervm_controller.go
dprince: Role is in!

controllers/controllervm_controller.go
dprince: I think I got it in the latest patch. It is an IP/netmask so I don't even need to use net.ParseCIDR format on it. Its ready to go straight from the ipset...

controllers/controllervm_controller.go
```diff
@@ -240,6 +267,39 @@ func (r *ControllerVMReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 	return ctrl.Result{}, nil
 }
 
+func (r *ControllerVMReconciler) setControllerHostStatus(instance *ospdirectorv1beta1.ControllerVM, hostname string, ipaddress string) {
```
dprince, line 270: Just realized that I added this function to set the ControllerHostStatus but I'm not actually calling it anywhere yet! On it...

controllers/baremetalset_controller.go
```diff
@@ -195,11 +194,12 @@ func (r *BaremetalSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
 
 	if op != controllerutil.OperationResultNone {
 		r.Log.Info(fmt.Sprintf("IPSet for %s successfully reconciled - operation: %s", instance.Name, string(op)))
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
```
dprince, line 197: We were missing these before. Important to have them in case it takes longer for the overcloudipsets to generate.

controllers/overcloudipset_controller.go
```diff
@@ -96,7 +96,7 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 
 	// iterate over the requested hostCount
 	for count := 0; count < instance.Spec.HostCount; count++ {
-		hostname := fmt.Sprintf("%s-%d", instance.Name, count)
+		hostname := fmt.Sprintf("%s-%d", instance.Spec.Role, count)
```
dprince, line 99: @stuggi @abays This makes it so we use the Role as the basis for the hostnames.

stuggi: yes, and we should make sure the role gets converted to lowercase as the role is case sensitive to construct the correct HostNameFormat for the deployment. I'll do that in my patch

stuggi (approved):

abays (approved):
/lgtm

PR #83, opened 2021-01-13
Move mandory deployment templates into the osp-director-operator

stuggi:


config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/openstack-k8s-operators/osp-director-operator
-￼ newTag: 0.0.1
+  newTag: 0.0.1
```
stuggi, line 8: there was some bad character and kustomize failed to set the tag when creating a new build

containers/provision_ip_discovery_agent/start.go
```diff
@@ -49,7 +49,12 @@ func init() {
 }
 
 func runStartCmd(cmd *cobra.Command, args []string) {
-	flag.Set("logtostderr", "true")
+	var err error
+	err = flag.Set("logtostderr", "true")
```
stuggi, line 53: fixed to pass new  ci checks. failed since flag.Set returns error with now check for it

pkg/overcloudipset/configmap.go
```diff
@@ -33,9 +35,20 @@ func CreateConfigMapParams(overcloudIPList ospdirectorv1beta1.OvercloudIPSetList
...
+
 	for _, ipset := range overcloudIPList.Items {
-		for count := 1; count <= ipset.Spec.HostCount; count++ {
-			hostname := fmt.Sprintf("%s-%d", ipset.Spec.Role, count)
+
+		hostnameFormat[fmt.Sprintf("%sHostnameFormat", ipset.Spec.Role)] = fmt.Sprintf("%s-%%index%%", strings.ToLower(ipset.Spec.Role))
+		roleCount[fmt.Sprintf("%sCount", ipset.Spec.Role)] = strconv.Itoa(ipset.Spec.HostCount)
+
+		for count := 0; count < ipset.Spec.HostCount; count++ {
+			hostname := fmt.Sprintf("%s-%d", strings.ToLower(ipset.Spec.Role), count)
```
stuggi, line 48: @dprince could common.CreateOrGetHostname just use the role as input instead of the hostKey using the cr name? then we could use it in all places where we have the roles available.

controllers/openstackclient_controller.go
```diff
@@ -93,26 +93,76 @@ func (r *OpenStackClientReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+	err = common.EnsureConfigMaps(r, instance, cms, &envVars)
+	if err != nil {
+		return ctrl.Result{}, nil
+	}
 
-	// check for required configMaps
-	configMaps := []string{
-		"tripleo-deploy-config",
-		"tripleo-deploy-sh",
+	// get tripleo-deploy-config, created/rendered by overcloudipset controller
```
abays, line 126: Maybe should say `tripleo-deploy-config-custom` instead

stuggi: yes, indeed

stuggi: done

pkg/overcloudipset/configmap.go
dprince: possibly. We may need to change how the data gets stored in the resulting Status block

dprince (approved):

PR #89, opened 2021-01-14
Add gateway to OvercloudNet

dprince:
The network info is now stored on each OvercloudIpset upon
creation.

Also, Updates baremetalset and controllervm to use the now
configurable but optional gateway on the ctlplane network.


controllers/baremetalset_controller.go
```diff
@@ -477,7 +477,7 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
 	templateParameters = make(map[string]interface{})
 	templateParameters["MgmtIp"] = ip.String()
 	templateParameters["MgmtInterface"] = instance.Spec.MgmtInterface
-	templateParameters["MgmtGateway"] = "192.168.25.1" //FIXME
+	templateParameters["MgmtGateway"] = ipset.Status.Networks["ctlplane"].Gateway
 	templateParameters["MgmtNetmask"] = fmt.Sprintf("%d.%d.%d.%d", netMask[0], netMask[1], netMask[2], netMask[3])
 	// TODO: Either generate the gateway from the network object or allow user specification via CR?
```
abays, line 482: We can remove this TODO now

controllers/overcloudipset_controller.go
```diff
@@ -106,12 +107,13 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 			if err != nil {
 				if k8s_errors.IsNotFound(err) {
 					r.Log.Info(fmt.Sprintf("OvercloudNet named %s not found!", netName))
-					return ctrl.Result{}, nil
 				}
 				// Error reading the object - requeue the request.
 				return ctrl.Result{}, err
 			}
 
+			instance.Status.Networks[network.Name] = network.Spec
```
abays, line 115: Why do we store an entire `OvercloudNet` spec here instead of just its name?  Just curious.

dprince: So the 'Spec' has the information we need in it, namely the gateway. I thought about creating a duplicate struct for this but the struct has what we need and as this is just a cache it seemed reasonable.

controllers/baremetalset_controller.go
dprince: Ack. will do

dprince: Fixed in the latest push.

templates/controllervm/cloudinit/networkdata
```diff
@@ -1,4 +1,7 @@
 version: 2
 ethernets:
   eth1:
-    addresses: [ {{ .ControllerIP }} ]
\ No newline at end of file
+    addresses: [ {{ .ControllerIP }} ]
+{{if .Gateway }}
+    {{ .Gateway }}
```
stuggi, line 6: this should be either gateway4: {{ .Gateway }} or gateway6: {{ .Gateway }}

https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v2.html#examples

stuggi: oh wait ignore that, you do that when setting templateParameters["Gateway"]

controllers/baremetalset_controller.go
```diff
@@ -477,9 +477,8 @@ func (r *BaremetalSetReconciler) baremetalHostProvision(instance *ospdirectorv1b
 	templateParameters = make(map[string]interface{})
 	templateParameters["MgmtIp"] = ip.String()
 	templateParameters["MgmtInterface"] = instance.Spec.MgmtInterface
-	templateParameters["MgmtGateway"] = "192.168.25.1" //FIXME
+	templateParameters["MgmtGateway"] = ipset.Status.Networks["ctlplane"].Gateway
```
stuggi, line 480: right now this is not really Mgmt, this is Ctl, should we rename all of this and follow up with adding Mgmt net implementation?

templates/controllervm/cloudinit/networkdata
dprince: Yeah, I could have implemented it differently, this way seemed cleaner.

controllers/baremetalset_controller.go
dprince: Yes. We had agreed to do this just got forgotted I think. Anyways I filed this: https://github.com/openstack-k8s-operators/osp-director-operator/issues/92

stuggi (approved):

PR #95, opened 2021-01-15
support customized root pwd

stuggi:
if PasswordSercret parameter is provided in the CR the controllers
set the root password via cloud init for the root user. It is expected
that the PasswordSercet data has a NodeRootPassword key with a plane
text password.

### DIFF
--- a/api/v1beta1/baremetalset_types.go
+++ b/api/v1beta1/baremetalset_types.go
@@ -42,6 +42,10 @@ type BaremetalSetSpec struct {
 	Networks []string `json:"networks"`
 	// Role the name of the Overcloud role this IPset is associated with. Used to generate hostnames.
 	Role string `json:"role"`
+	// PasswordSecret the name of the secret used to optionally set the root pwd by adding
+	// NodeRootPassword: <base64 enc pwd>
+	// to the secret data
+	PasswordSecret string `json:"passwordSecret,omitempty"`
 }
 
 // BaremetalSetStatus defines the observed state of BaremetalSet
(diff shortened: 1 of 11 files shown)


api/v1beta1/baremetalset_types.go
```diff
@@ -42,6 +42,8 @@ type BaremetalSetSpec struct {
 	Networks []string `json:"networks"`
 	// Role the name of the Overcloud role this IPset is associated with. Used to generate hostnames.
 	Role string `json:"role"`
+	// PasswordSecret used to e.g specify root pwd
```
dprince, line 45: Should we add a short version of the expected ConfigMap contents to this comment so that the end user will easily see it. If we do this it should show up as part of the CRD which is ideal...

api/v1beta1/controllervm_types.go
```diff
@@ -42,6 +42,8 @@ type ControllerVMSpec struct {
 	Networks []string `json:"networks"`
 	// Role the name of the Overcloud role this IPset is associated with. Used to generate hostnames.
 	Role string `json:"role"`
+	// PasswordSecret used to e.g specify root pwd
```
dprince, line 45: ditto here.

api/v1beta1/baremetalset_types.go
stuggi: done

api/v1beta1/controllervm_types.go
stuggi: done

dprince:
@stuggi I think just a 'make manifests' and we are ready to go now

dprince (approved):

PR #110, opened 2021-01-21
Validation webhooks

abays:
WARNING: Lacks proper OLM support.  If we decide to merge this, the new `master` branch must be built and pushed to `quay.io/openstack-k8s-operators` before merging and using https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/47!

...

I have compared the bundle data of an existing operator (containing webhooks) with the bundle data that is generated by this patch (that is, generated by `./build_and_push_images.sh`).  Here are the comparison operator's bundle manifests:

https://github.com/rojkov/community-operators/tree/intel-operator/upstream-community-operators/intel-device-plugins-operator/0.19.0

You will notice that it only contains CRDs and the CSV.  If you look more closely, you will see that its CSV has webhooks defined (near the bottom) just as ours does when you generate the bundle.  Aside from some port field/values that our CSV lacks in the webhooks section, I don't see a difference.  Yet, ostensibly, this comparison operator deploys properly via OLM (or did, in an earlier version of k8s) while ours does not.

What currently happens when you deploy this patch's operator via OLM is this:

1. The controller-manager deployment *does* have a webhook webserver port and volume/mount included.
2. The volume referenced in step 1 is a `Secret` (containing TLS data) that is *supposed* to be created by `cert-manager` via a `Certificate` resource.  This `Certificate` is, in theory, something that OLM should be generating during CSV deployment.
3. The `Certificate` referenced in step 3 is *not* created by OLM (nor is the `Issuer` resource that is associated with it).
4. The `Service` that exposes the controller manager's webhook webserver to the cluster API is *not* created by OLM during deployment, though it allegedly should in fact be doing so.
5. The `ValidationWebhookConfiguration` represented at the bottom of the CSV is *not* created by OLM during deployment, though it supposedly should be.

As a result of the issues listed above, the controller manager pod gets stuck in the `ContainerCreating` stage, as it waits forever trying to mount the `Certificate`-generated `Secret` that would contain the TLS data.


abays:
Closing in favor of https://github.com/openstack-k8s-operators/osp-director-operator/pull/119

PR #117, opened 2021-01-27
Add baseImageVolumeName

dprince:
This patch adds baseImageVolumeName and removes imageImportStorageClass.
By allowing an external baseImageVolume to be specified we allow
end users to pre-create their own DataVolume in any manner they choose
thus optimizing base image imports for their environments.

Specifying a baseImageVolume will effectively disable the baseImageURL
param although baseImageURL is still left as required for now
until we get validation hooks enabled to enforce either/or
validation across these two parameters.


stuggi (approved):
/lgtm

PR #119, opened 2021-01-27
Self-deploying webhooks

abays:
Since OLM-based webhook deployment is not currently working for us, this is a patch that enables webhooks by:

1. Injecting a `Service` into the operator bundle to allow traffic to the operator's webhook server.
2. Using the operator's manager's initialization to create any validation webhooks we require.
3. *Attempting* to clean-up webhook definitions (`ValidatingWebhookConfiguration`) when the operator manager pod is deleted.

The rest of the webhook functionality added by this patch is otherwise valid and is currently best-practice as far as `operator-sdk`-based webhooks are concerned.  Once the issues enumerated in https://bugzilla.redhat.com/show_bug.cgi?id=1921000 are resolved, the three workarounds can be removed.  We will, however, keep the rest.

NOTE: When running the operator locally, you will need to use:

```
WATCH_NAMESPACE=openstack,openshift-machine-api ENABLE_WEBHOOKS=false make run
```

...this is a standard approach and is not due to any of the three workarounds listed above.  Of course, in doing so, the webhooks will not be running.


abays:
Yet another close...this time in favor of: https://github.com/openstack-k8s-operators/osp-director-operator/pull/126

PR #121, opened 2021-01-28
Use lowercase role names for host names

abays:
Fixes an issue where VMSet generated infinite host names and empty IP addresses by using VMSet role names (plus an index) as the "host key" for VMSet host name generation.  To reproduce this beforehand, create a VMSet with differing instance name and role name (i.e. osp-database and database, respectively).


stuggi (approved):
/lgtm

PR #126, opened 2021-02-01
Webhooks with OLM deployment

abays:
Admission webhook support with proper deployment and cleanup via OLM


bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=osp-director-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.2.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.1.0
```
dprince, line 8: Remove this?

config/manifests/bases/osp-director-operator.clusterserviceversion.yaml
```diff
@@ -4,9 +4,9 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
-    operators.operatorframework.io/builder: operator-sdk-v1.2.0
-    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
+    operators.operatorframework.io/builder: operator-sdk-v1.1.0
```
dprince, line 7: I don't think there should be changes to this file either right?

config/webhook/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 resources:
 - manifests.yaml
-- service.yaml
+#- service.yaml
```
dprince, line 3: why is this commented out?

scripts/build_and_push_images.sh
```diff
@@ -4,7 +4,7 @@ set -ex
 #
 # Builds and pushes operator, bundle and index images for a given version of an operator
 #
-# NOTE: Requires opm and operator-sdk to be installed!
+# NOTE: Requires yq, make, podman, opm and operator-sdk to be installed!
```
dprince, line 7: we don't require 'yq' again do we (I don't have it!)

config/webhook/kustomization.yaml
abays: It's not used, so I commented it out.  OLM generates the service at deploy-time automagically

config/manifests/bases/osp-director-operator.clusterserviceversion.yaml
abays: Ack

scripts/build_and_push_images.sh
abays: Ack

bundle.Dockerfile
abays: Ack

PR #133, opened 2021-02-05
Integrate vip into ipam

stuggi:
Integration of VIP into IPAM, also:
* moves to use a AssignIPDetails struct with all details to
  reduce the parameter list of the used functions
* alligns the CRD status of vmset, overcloud and openstackclient.
  TODO for baremetalset when adding isolated networks architecture.
* moves the deployed port map and the predictible ip pool rendering
  to an extrenal template file
* prepares for isolated network to add resource definitions to
  predictibale ips template


controllers/controlplane_controller.go
```diff
@@ -139,6 +139,52 @@ func (r *ControlPlaneReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+		AddToPredictableIPs: false,
+	}
+	ipset, op, err := common.OvercloudipsetCreateOrUpdate(r, instance, ipsetDetails)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if op != controllerutil.OperationResultNone {
+		r.Log.Info(fmt.Sprintf("IPSet for %s successfully reconciled - operation: %s", instance.Name, string(op)))
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
```
bogdando, line 156: shouldn't the success path be different to the pending replicas case?..

controllers/vmset_controller.go
```diff
@@ -306,11 +313,23 @@ func (r *VMSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-func (r *VMSetReconciler) setVMHostStatus(instance *ospdirectorv1beta1.VMSet, hostname string, ipaddress string) {
-	// Set status for vmHosts
-	instance.Status.VMHosts[hostname] = ospdirectorv1beta1.VMHostStatus{
-		Hostname:  hostname,
-		IPAddress: ipaddress,
+func (r *VMSetReconciler) setNetStatus(instance *ospdirectorv1beta1.VMSet, hostnameDetails *common.Hostname, netName string, ipaddress string) {
+
+	// If VMSet status map is nil, create it
+	if instance.Status.VMHosts == nil {
+		instance.Status.VMHosts = map[string]ospdirectorv1beta1.HostStatus{}
```
bogdando, line 320: shouldn't this use make()?

controllers/controlplane_controller.go
stuggi: Not sure I understand your comment. OperationResultNone [1] means that the resource has not been changed. 
But I think we could remove the reconcile here and just log the done action. It does not hurt, but it seems to be a not necessary reconcile 

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#OperationResult

controllers/controlplane_controller.go
```diff
@@ -139,6 +138,49 @@ func (r *ControlPlaneReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error
...
+	ipset, op, err := common.OvercloudipsetCreateOrUpdate(r, instance, ipsetDetails)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if op != controllerutil.OperationResultNone {
+		r.Log.Info(fmt.Sprintf("IPSet for %s successfully reconciled - operation: %s", instance.Name, string(op)))
+	}
+
+	if len(ipset.Status.HostIPs) != controlplane.Count {
```
abays, line 157: I've seen this IP vs count `!=` check cause problems on scale-down for `BaremetalSet` and `VMSet`.  Switching to `<` seemed to fix the isssue.  Might need that here as well, unless the IPs are in fact being removed from the `OvercloudIPSet` during scale-down.

controllers/openstackclient_controller.go
```diff
@@ -122,15 +122,26 @@ func (r *OpenStackClientReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	r.Log.Info(fmt.Sprintf("OpenStackClient hostname set to %s", hostname))
+	r.Log.Info(fmt.Sprintf("OpenStackClient %s hostname set to %s", hostnameDetails.IDKey, hostnameDetails.Hostname))
+
+	// TODO: maschuppert - atm the openstackclient only has the ctlplane network.
+	//       for debugging it probably makes sense to attach the openstackclient to all networks?
+
+	// update VIP network status
```
abays, line 139: Is this actually VIP network status, given `VIP: false` on line 128?

controllers/overcloudipset_controller.go
```diff
@@ -142,7 +143,22 @@ func (r *OvercloudIPSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
...
+					RangeStart:          start,
+					RangeEnd:            end,
+					Reservelist:         network.Status.Reservations,
+					ExcludeRanges:       []string{},
+					IDKey:               fmt.Sprintf("%s-%d", strings.ToLower(instance.Spec.Role), count),
+					Hostname:            hostname,
+					VIP:                 instance.Spec.VIP,
+					Role:                instance.Spec.Role,
+					AddToPredictableIPs: instance.Spec.AddToPredictableIPs,
+				})
```
abays, line 161: Can we just use `hostname` as the `IDKey` here?  The func that generates the hostnames uses `Role`+`-<lowest index available>`, and it will pick gaps in the middle when they're available (to address the concern in line 149).  I think I might be missing the larger picture, though, for why we're adding `IDKey`.

abays: Based on your comments in your changes to `pkg/common/hostname.go`, I think there is something I am not seeing that would provide the context I need to understand why the answer to my question above might be "no". :-)

controllers/controlplane_controller.go
stuggi, line 157: It probably won't be an issue here as the controlplane.Count can only be 1 or 0, but lets keep it the same as in the other controllers.

controllers/openstackclient_controller.go
stuggi: yes,  it is not. previously I had openstackclient VIP: true as there might be always just a single openstackclient, but switched it to false later and missed to update the comment here, thanks!

controllers/overcloudipset_controller.go
stuggi: The idea of having IDKey is related to [1] to never reuse an IDKey when a host gets deleted. We need a way to always re/create ips-from-pool-all from scratch with the information of deleted hosts(marked as deleted):

```
Even if an overcloud node is deleted, its entry in the IP lists should not be removed. The IP list is based on the underlying Heat indices, which do not change even if nodes are deleted. To indicate that a given entry in the list is no longer used, the IP value can be replaced with a value such as “DELETED” or “UNUSED”.

In short, entries should never be removed from the IP lists, only changed or added.
```

Also having IDKey we have the chance, if we ever decide, to provide a way to set custom hostnames (compute-0 -> mysupercompute [2])and it won't conflict with creating ips-from-pool-all.

For the above we don't have an implementation in hostname.go yet, using IDKey here is just a pre-requisit. If we say
* hostnames are never being re-used
* index always grows, like we had cmp1-10, scaled down to cmp9, scaled up to 10 again, but the new 10th cmp get cmp11 as hostname 
* keep the old (removed) hostnames to track the status in ips-from-pool-all

then we could just switch to use hostnames. Hope that makes sense :)

[1] https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/provisioning/node_placement.html#predictable-ips

[2] https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/provisioning/node_placement.html#custom-hostnames

stuggi:
urgs .. did PR response in vscode and it messed up the comments :)

controllers/openstackclient_controller.go
stuggi: updated

controllers/controlplane_controller.go
stuggi: changed

stuggi: changed

controllers/overcloudipset_controller.go
dprince: I had the same thoughts as @abays with regards to IDKey (like could we just use hostname). 

And thanks @stuggi for the explanation. Perhaps we keep IdKey for now, but I wonder is there a way to hide it on the CLI or something to keep it more of an internal resource. It does look odd on the Status when you look at the description of an overcloudnet.

PR #140, opened 2021-02-09
Make VMSet baseImageURL optional

abays:
Makes VMSet `baseImageURL` and `baseImageVolumeName` optional, but requires one of them to be set


stuggi (approved):
/lgtm

PR #142, opened 2021-02-10
Prevent unsanctioned VM deletes

abays:
Adds finalizers to VMSet VM resources to prevent users from manually deleting them.


config/default/manager_webhook_patch.yaml
```diff
@@ -12,3 +12,7 @@ spec:
         - containerPort: 4343
           name: webhook-server
           protocol: TCP
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/osp-director-operator-finalizer-clean", "start"]
```
stuggi, line 18: does that run also on "normal" operator stop/start, like on updates? don't think we want the finalizers to be removed on day2 operations like updates.

controllers/vmset_controller.go
```diff
@@ -109,6 +109,52 @@ func (r *VMSetReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
+	// What VMs do we currently have for this VMSet?
+	virtualMachineList, err := common.GetAllVirtualMachinesWithLabel(r, map[string]string{
+		OwnerNameLabelSelector: instance.Name,
```
stuggi, line 114: not required for this change, but we probably want to add a role label to the VMs and filter here on the role specific VMs which is needed when we start to support split composable roles into different vmsets

config/default/manager_webhook_patch.yaml
abays: Good point.  Yes, it would indeed.  Once the updated operator pod re-appeared, though, the finalizers would be put back in place.  This does, of course, open a window where the user could skirt our deletion protection.

controllers/vmset_controller.go
abays: Would the uniqueness of `instance.Name` + `instance.Namespace` not be enough to ensure the VMs are filtered properly?  Or are you saying elsewhere where we might be filtering VMs and might not have the instance name/namespace available?

stuggi: yes it should, you are right.

PR #151, opened 2021-02-17
Use tripleoclient from tripleo

stuggi:
[1] adds the changes to tripleoclient so that we can use the tripleo
container image instead of our curr current created. This allign
cloud-admin uid with tripleoclient image form [1].

[1] https://review.opendev.org/c/openstack/tripleo-common/+/776174


abays (approved):
/lgtm

PR #152, opened 2021-02-18
Rename BaremetalSet to OpenStackBaremetalSet

dprince:


bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=osp-director-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.2.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.4.0
```
abays, line 8: Is this the SDK version we should be using now?  It's good with me, just curious.

config/crd/bases/osp-director.openstack.org_openstackbaremetalsets.yaml
```diff
@@ -147,9 +147,8 @@ spec:
                 secret data'
               type: string
             provisionServerName:
-              description: Name of the pre-created ProvisionServer to use. If su ProvisionServerName
-                Optional. If supplied will be used as the base Image for the baremetalset
-                instead of RhelImageURL.
+              description: ProvisionServerName Optional. If supplied will be used
+                as the base Image for the baremetalset instead of RhelImageURL.
```
abays, line 151: Nit: Here and elsewhere in comments, should `baremetalset`/`BaremetalSet` also be renamed?  I think it's pretty clear what this refers to in the context, so I could go either way.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/osp-director-operator
-  newTag: 0.0.1
+  newName: quay.io/dprince/osp-director-operator
```
abays, line 7: Should probably undo this

config/webhook/manifests.yaml
```diff
@@ -23,15 +23,15 @@ webhooks:
...
 - clientConfig:
     caBundle: Cg==
     service:
       name: webhook-service
       namespace: system
-      path: /validate-osp-director-openstack-org-v1beta1-controlplane
+      path: /validate-osp-director-openstack-org-v1beta1-baremetalset
   failurePolicy: Fail
-  name: vcontrolplane.kb.io
+  name: vbaremetalset.kb.io
```
abays, line 34: Might need to change this to the new name

dprince: ack, this file is generate so what I missed is the changing the annotation above. Will fix...

dprince: @abays fixed in latest commit

config/manager/kustomization.yaml
dprince: fixed in latest commit

PR #155, opened 2021-02-18
Add Kuttl test for OpenStackControlPlane scale up/down

abays:
My apologies for the size of this PR, but there's really no way around it if we want to use `kuttl` tests.  Perhaps it is best to clone my fork/branch and follow the instructions in the `kuttl-test.yaml` to run these tests and verify that they work in your environment.  Then if that succeeds, proceed to review this PR.

Also, in using project-wide find-and-replace for certain case-sensitive CRD names (to rebase against our CRD renaming PRs), files were modified that otherwise wouldn't have been.  This increased the `Files changed` significantly, but these extras should be quick to eyeball and move beyond.

And finally a note on how the `kuttl` YAMLs work: 

- The `XX-` prefixes denote the index of the step involved.  So each test begins with `00` and goes to whatever is required
- If there is an `XX-assert.yaml` for a given `XX-` index, that is declaring what the final state of that step should be in terms of what _should_ exist
- If there is an `XX-errors.yaml` for a given `XX-` index, that is declaring what the final of that step should have in terms of what _should not_ exist


abays:
Rebasing made this a mess.  Going to recreate.

PR #158, opened 2021-02-22
OpenStackControlPlane scale up/down kuttl test

abays:
My apologies for the size of this PR, but there's really no way around it if we want to use `kuttl` tests.  Perhaps it is best to clone my fork/branch and follow the instructions in the `kuttl-test.yaml` to run these tests and verify that they work in your environment.  Then if that succeeds, proceed to review this PR.

Also, in using project-wide find-and-replace for certain case-sensitive CRD names (to rebase against our CRD renaming PRs), files were modified that otherwise wouldn't have been.  This increased the `Files changed` significantly, but these extras should be quick to eyeball and move beyond.

And finally a note on how the `kuttl` YAMLs work: 

- The `XX-` prefixes denote the index of the step involved.  So each test begins with `00` and goes to whatever is required
- If there is an `XX-assert.yaml` for a given `XX-` index, that is declaring what the final state of that step should be in terms of what _should_ exist
- If there is an `XX-errors.yaml` for a given `XX-` index, that is declaring what the final state of that step should have in terms of what _should not_ exist


config/samples/osp-director_v1beta1_openstackcontrolplane.yaml
```diff
@@ -7,13 +7,13 @@ spec:
...
     Role: Controller
-    controllerCount: 1
+    controllerCount: 0
     cores: 2
     memory: 20
     diskSize: 50
     baseImageURL: http://download.eng.brq.redhat.com/brewroot/packages/rhel-guest-image/8.3/417/images/rhel-guest-image-8.3-417.x86_64.qcow2
     storageClass: host-nfs-storageclass
-    #baseImageVolumeName: #optional
+    baseImageVolumeName: openstack-base-img #optional
```
dprince, line 16: This will change the default OperatorHub example for this CR. Specifically this gets injected via alm-examples into the resulting CSV when we built the operator... so I'm not sure we'd want both baseImageURL and baseImageVolumeName specified as if that is the case only baseImageVolumeName gets applied.

abays: Don't worry, I have that covered by `kustomize` patches that are applied when the CSV is generated

kuttl-test.yaml
```diff
@@ -0,0 +1,35 @@
...
+#    - mv kubectl-kuttl_0.9.0_linux_x86_64 /usr/local/bin/kubectl-kuttl
+#    - chmod 755 /usr/local/bin/kubectl-kuttl
+# 2. An OCP 4.6+ cluster with OSP Director Operator has been deployed via https://github.com/openstack-k8s-operators/osp-director-dev-tools
+# 3. CLI user has access to $KUBECONFIG
+# 4. There are preferrably NO existing OSP Director Operator CRs deployed
+#    - dev-tools probably installed them, so remove via:
+#        oc delete baremetalset --all
+#        oc delete controlplane --all
+#        oc delete vmset --all
+#        oc delete overcloudnet --all
```
stuggi, line 19: this is now:
openstackbaremetalset
openstackcontrolplane
openstackvmset
openstacknet

abays: Ack

dprince (approved):
Very nice job putting this together. For some reason my bundle image isn't building correctly locally so I've not been able to test it. (didn't get to the bottom of this, but I see its passing CI so perhaps just in my env). I do wonder about the edits to the sample CRs, but reusing those also makes some sense. We may need multiple version of them anyway though right to be able to test alternate scenerios though right?

PR #161, opened 2021-02-23
Remove long names for now

abays:
Just undoing certain prov server and baremetal set identifier name changes until we decide on short names


stuggi (approved):
/lgtm

PR #162, opened 2021-02-23
Introduce composable roles support to split roles into multiple VMsets

stuggi:
Changes the OpenstackControlPlane CRD to be able to specify a list of
VirtualMachineRoles which can be either mapped to TripleO roles, or
used to setup virtual machines to host 3rd party services which won't
get installed using TripleO. A new parameter isTripleoRole got added
to indicate if a role is managed via TripleO and its information gets
rendered into the TripleO environment files.
E.g:
~~~
  virtualMachineRoles:
  - role: Controller
    isTripleoRole: true
    controllerCount: 1
    cores: 2
    memory: 20
    diskSize: 50
    baseImageURL: http://.../image.qcow2
    storageClass: host-nfs-storageclass
    networks:
    - ctlplane
    ospNetwork:
      bridgeName: br-osp
      name: osp
      desiredState:
        interfaces:
        - name: br-osp # must be same as bridgeName
          description: Linux bridge with enp6s0 as a port
          type: linux-bridge
          state: up
          bridge:
            options:
              stp:
                enabled: false
            port:
            - name: enp6s0
  - role: SomeCustomVM
    isTripleoRole: false
    controllerCount: 1
    cores: 2
    memory: 20
    diskSize: 50
    baseImageURL: http://.../image.qcow2
    storageClass: host-nfs-storageclass
    networks:
    - ctlplane
    ospNetwork:
      bridgeName: br-osp
      name: osp
      desiredState:
        interfaces:
        - name: br-osp # must be same as bridgeName
          description: Linux bridge with enp6s0 as a port
          type: linux-bridge
          state: up
          bridge:
            options:
              stp:
                enabled: false
            port:
            - name: enp6s0
~~~


api/v1beta1/openstackvmset_types.go
```diff
@@ -42,8 +42,10 @@ type OpenStackVMSetSpec struct {
 	OSPNetwork Network `json:"ospNetwork"`
 	// Networks the name(s) of the OpenStackNetworks used to generate IPs
 	Networks []string `json:"networks"`
-	// Role the name of the Overcloud role this IPset is associated with. Used to generate hostnames.
+	// Role the name of the Overcloud role this RoleSpec is associated with. If it is a TripleO role, the name must match.
 	Role string `json:"role"`
+	// in case of external functionality, like 3rd party network controllers, set to false to ignore role in rendered overcloud templates.
+	IsTripleoRole bool `json:"isTripleoRole"`
```
abays, line 48: Could we use `kubebuilder` annotations to set a default of "true" and then have this be `omitempty`?  I'm just guessing that most of the time this will be set to true?

dprince, line 48: Instead of IsTripleoRole we might consider calling this ValidateRole or something more generic.

dprince:
In general I think this looks great. Just a question about the parametername which we might consider (not asking to change it yet...)

api/v1beta1/openstackvmset_types.go
stuggi: yes that makes send, but to be able to use default values we need to switch to apiextensions.k8s.io/v1, default values can not be set using current apiextensions.k8s.io/v1beta1

~~~
Forbidden: must not be set (cannot set default values in apiextensions.k8s.io/v1beta1 CRDs, must use apiextensions.k8s.io/v1)
~~~

We can switch to that setting CRD_OPTIONS ?= "crd:crdVersions=v1" in the Makefile, but have to verify if it won't break anything else. In general I think this is be a good thing as we could set decent default values for our parameters.

https://github.com/operator-framework/operator-sdk/issues/3924
https://github.com/operator-framework/operator-sdk/issues/3580

stuggi: we can find a better name for it. the main reason for this is to prevent 3rd party roles to be rendered into the tripleo environment files, like per deployed server port map or role count.

stuggi: switch to v1 also sets preserveUnknownFields: false , which is good imo as it prevents accept CRs with typos in its name and fails early.

[1] https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#pruning-don-t-preserve-unknown-fields

api/v1beta1/openstackcontrolplane_types.go
```diff
@@ -22,35 +22,39 @@ import (
 
 // OpenStackControlPlaneSpec defines the desired state of OpenStackControlPlane
 type OpenStackControlPlaneSpec struct {
-	Controller OpenStackControllerSpec `json:"controller"`
+	// List of VirtualMachine roles
+	VirtualMachineRoles []OpenStackVirtualMachineRoleSpec `json:"virtualMachineRoles"`
```
stuggi, line 26: maybe we should use here a map instead of an array to be able to easy patch specific roles for kuttl tests when testing multiple VM roles

api/v1beta1/openstackvmset_types.go
stuggi: done, I added a commit to move to apiextensions.k8s.io/v1, also set a default value for this boolena

api/v1beta1/openstackcontrolplane_types.go
stuggi: done with latest commit

kuttl-test.yaml
```diff
@@ -28,7 +28,6 @@ testDirs:                      # The actual kuttl tests
   - ./tests/kuttl/tests
 crdDir: ./config/crd/bases     # CRDs needed for our operator
 manifestDirs:                  # YAMLs to apply before each test
```
abays, line 30: Is it okay to have this key without a value?

tests/kuttl/tests/openstackcontrolplane_scale/06-assert.yaml
```diff
@@ -1,43 +0,0 @@
-#
-# Check for:
-#
-# - OpenStackNet (should be preserved with all historical reservations)
-#
```
abays, line 5: Is this no longer a valid test?

stuggi: it is! I deleted 06 by accident and only restored 06-delete_openstackcontrolplane.yaml and 06-errors.yaml .. will restore it.

kuttl-test.yaml
stuggi: it seems kuttl handles that, but yeah the clean way would be to set it to {}

abays: Ah, no worries then if kuttl doesn't complain

tests/kuttl/tests/openstackcontrolplane_scale/06-assert.yaml
stuggi: done

kuttl-test.yaml
stuggi: done

config/samples/osp-director_v1beta1_openstackcontrolplane_with_additional_custom_role.yaml
```diff
@@ -0,0 +1,49 @@
...
+      memory: 20
+      diskSize: 50
+      baseImageURL: http://download.eng.brq.redhat.com/brewroot/packages/rhel-guest-image/8.3/417/images/rhel-guest-image-8.3-417.x86_64.qcow2
+      storageClass: host-nfs-storageclass
+      baseImageVolumeName: openstack-base-img #optional
+      networks:
+      - ctlplane
+    somecustomrole:
+      roleName: SomeCustomRole
+      isTripleoRole: false
```
dprince, line 40: This is renamed now right?

dprince (changes requested):

config/samples/osp-director_v1beta1_openstackcontrolplane_with_additional_custom_role.yaml
stuggi: no, but we default to true, so you only have to set it on non tripleo roles.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/11c5c1205a83[…]d4a78e85152bd0f8389d/api/v1beta1/openstackcontrolplane_types.go

PR #169, opened 2021-03-02
Change CRD patches to v1

abays:
The patch files in `config/crd/patches` were still targeting `v1beta` CRDs, and need to target `v1` now instead


stuggi (approved):
/lgtm

PR #171, opened 2021-03-04
Fix golangci github action

mpryc:
Pin golangci-lint to the latest stable version and increase
default timeout which was 1m.


stuggi (approved):
/lgtm

PR #172, opened 2021-03-04
SRIOV integration for OpenStackVmSets

abays:
NOTE: This will require changes to dev-tools before it should ever be merged!  Among them:

- Logic to install SRIOV network operator
- Changes to OLM subscription to watch `openshift-sriov-network-operator` namespace in addition to those already included
- For dev/testing in virtualized environments: logic to use the new "virtualized SRIOV" functionality

Adds a new `OspNetwork` option to `OpenStackVmSets` (which will we later surface in `OpenStackControlPlane` once we finalize the design) to allow a user to configure SRIOV on CNV VMs.


abays:
This is now outdated.  We are taking a different approach.

PR #175, opened 2021-03-10
Switch away from bindata for vm resources

stuggi:
This switch to use the schema for NetworkConfigurationPolicy,
NetworkAttachmentDefinition and VirtualMachine instead of the
previous used bindata templates.

For NetworkConfigurationPolicy and NetworkAttachmentDefinition
to use dynamic client + ssa as CreateOrUpdate can not be used
because NetworkConfigurationPolicy is non namespaced.

This just changes the current implementation to use the above.
This is reqired to make it easier to enabled support for
network isolation where a dynamic list of
NetworkAttachmentDefinition needs to be created depending on
the used OSP network list.

With this change it is now also possible to modify vm resources
by changing the CR parameters, e.g. it is now possible to change
configuration parameters, like CPU/memory/... for the controllers.
When changed the controller pod needs to be recreated.
(virtctl stop/start controller-0)


config/samples/osp-director_v1beta1_openstackclient.yaml
```diff
@@ -6,5 +6,23 @@ metadata:
 spec:
   imageURL: quay.io/openstack-k8s-operators/tripleo-deploy
   deploymentSSHSecret: osp-controlplane-ssh-keys
+  ospNetwork:
+    name: osp
+    bridgeName: br-osp
```
abays, line 11: One thing I've been wondering as I've worked on the SRIOV integration is what happens when this common `Network` type (whether here in `OpenStackClient` or in `OpenStackVmSet`) conflicts based on `name`, `bridgeName` and `desiredState`.  What I mean is: what if some used the same `name` and `bridgeName` across two definitions (and the same `nodeSelector`, I suppose), but they had differing `desiredState`?  If this is a problem, should we maybe add webhooks to validate?

stuggi: yes that as a valid point and we should add it. we could do it in a follow up. this PR is intended to provide 1:1 transition from what we have atm.

abays: I agree.  Not needed in this PR.  I will add a task for it to the project.

abays (approved):
Nice work.  Your approach to using the dynamic client seems better than what I did for the provisioning server agent.  Perhaps we need to update that to work in the same manner.

dprince (approved):

PR #178, opened 2021-03-11
Small bindata + net-att-def cleanup

abays:
Removes bindata from Dockerfile and temporarily moves net-attach-def perms to cluster-scope for the meantime

### DIFF
--- a/controllers/openstackvmset_controller.go
+++ b/controllers/openstackvmset_controller.go
@@ -81,7 +81,8 @@ func (r *OpenStackVMSetReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=core,resources=pods;persistentvolumeclaims;events;configmaps;secrets,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=cdi.kubevirt.io,namespace=openstack,resources=datavolumes,verbs=create;delete;get;list;patch;update;watch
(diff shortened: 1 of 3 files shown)


controllers/openstackvmset_controller.go
```diff
@@ -81,7 +81,8 @@ func (r *OpenStackVMSetReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=core,resources=pods;persistentvolumeclaims;events;configmaps;secrets,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=cdi.kubevirt.io,namespace=openstack,resources=datavolumes,verbs=create;delete;get;list;patch;update;watch
-// +kubebuilder:rbac:groups=k8s.cni.cncf.io,namespace=openstack,resources=network-attachment-definitions,verbs=create;delete;get;list;patch;update;watch
+// FIXME: Temporarily switching next annotation to cluster-scope, as controller is watching for these in other namespaces
+// +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=create;delete;get;list;patch;update;watch
```
stuggi, line 85: I don't think we need that, we only create the network-attachment-definition in the openstack namespace, no?

abays: With just openstack, the controller was throwing an error saying that it didn't have permission to watch `net-att-def` in the `openshift-machine-api` namespace.  I've seen this happen with other resources beforehand.  As a band-aid, I just switched the perms to cluster-scope, but yeah, we need some proper investigation here.

stuggi: ah right, because we have not watched them before.

PR #179, opened 2021-03-12
Check nad CNIconfig

stuggi:
Depends on: #178


abays:
The check failure is just because this branch doesn't have #178 

/lgtm

PR #180, opened 2021-03-16
Integrate nncp and nad managment into ocnet controller

stuggi:
ATM OSPnetwork attach configuration is handled via the vmset
controller. The networks handled by the osnet controller are
the OSP networks, therefore the nncp and nad configuration to
each network should be handled by the osnet controller.

PR #184, opened 2021-03-22
Support tripleo network isolation using vlan

stuggi:
To attach the controller VM to the different VLANs, for each
VLAN a network-attachment-definition gets created, which was
done via the refactoring of the OSPnetwork in [1].

Adds optional OpenStackClientNetworks list parameter to the
OpenStackControlPlaneSpec CRD to allow customization of the
networks the OpenStackClient pod gets attached. Default is
ctlplane and external. For debugging it can be usefull to be
able to attach a openstackclient pod to any of the created
networks.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/pull/180


abays (commented):
Nice work.  I'm good with merging this once @dprince has a look.

stuggi:
I have found one issue, need to investigate

api/v1beta1/openstackcontrolplane_types.go
```diff
@@ -28,6 +28,11 @@ type OpenStackControlPlaneSpec struct {
 	OpenStackClientImageURL string `json:"openStackClientImageURL"`
 	// PasswordSecret used to e.g specify root pwd
 	PasswordSecret string `json:"passwordSecret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={ctlplane,external}
+	// OpenStackClientNetworks the name(s) of the OpenStackClientNetworks used to attach the openstackclient to
+	OpenStackClientNetworks []string `json:"openStackClientNetworks"`
```
dprince, line 35: I wonder if unset if we just always made these the same as the 'controlplane' vmset networks?

dprince (approved):

api/v1beta1/openstackcontrolplane_types.go
stuggi: yes we could do that if we want the openstackclient to always be attached to all the networks

PR #191, opened 2021-03-23
Adjust Dockerfile to be osbs friendly.

mpryc:
To automatically create OSBS compatible Dockerfile both upstream
and downstream should be similar and controlled by args that
are different.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,40 +1,65 @@
-ARG GOLANG_BUILDER=golang:1.13
+ARG GOLANG_BUILDER=golang:1.14
 ARG OPERATOR_BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 # Build the manager binary
 FROM ${GOLANG_BUILDER} AS builder
 
-ARG GO_BUILD_EXTRA_ARGS
+#Arguments required by OSBS build system
+ARG CACHITO_ENV_FILE=/remote-source/cachito.env
 
-WORKDIR /workspace
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
+ARG REMOTE_SOURCE=.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,40 +1,65 @@
...
 WORKDIR /
-COPY --from=builder /workspace/manager .
-COPY --from=builder /workspace/templates /usr/share/osp-director-operator/templates/.
-USER nonroot:nonroot
+
+# Install operator binary to WORKDIR
+COPY --from=builder ${DEST_ROOT}/manager .
+
+# Install templates
+COPY --from=builder ${DEST_ROOT}/templates ${OPERATOR_TEMPLATES}
```
yazug, line 61: does this work even if it is missing the subdirectory?

Dockerfile
```diff
@@ -1,40 +1,65 @@
...
 FROM ${OPERATOR_BASE_IMAGE}
 
+ARG DEST_ROOT=/dest-root
+ARG USER_ID=nonroot:nonroot
+
+LABEL   com.redhat.component="osp-director-operator-container" \
+        name="osp-director-operator" \
+        version="1.0" \
+        summary="OSP Director Operator" \
+        io.k8s.name="nova-operator" \
```
yazug, line 47: copy paste pasta mixup, name change didn't happen here.

Dockerfile
mpryc: Yes,
https://docs.docker.com/engine/reference/builder/#copy

If <dest> doesn’t exist, it is created along with all missing directories in its path.

Dockerfile
mpryc: ?

mpryc: ok got it s/nova/osp-director

stuggi (approved):
/lgtm

PR #195, opened 2021-03-24
Remove hard coded OpenStackClient storageClass

stuggi:
Adds OpenStackClientStorageClass parameter to the controlplane
CRD to be able to controll the storage class the pvc for the
openstackclient gets created.

### DIFF
--- a/api/v1beta1/openstackclient_types.go
+++ b/api/v1beta1/openstackclient_types.go
@@ -35,6 +35,9 @@ type OpenStackClientSpec struct {
 	// +kubebuilder:default={ctlplane,external}
 	// Networks the name(s) of the OpenStackNetworks used to generate IPs
 	Networks []string `json:"networks"`
+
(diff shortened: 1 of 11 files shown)


api/v1beta1/openstackclient_types.go
```diff
@@ -35,6 +35,9 @@ type OpenStackClientSpec struct {
 	// +kubebuilder:default={ctlplane,external}
 	// Networks the name(s) of the OpenStackNetworks used to generate IPs
 	Networks []string `json:"networks"`
+
+	// StorageClass to be used for the controller disks
```
abays, line 39: Maybe just to avoid confusion, this should say "client" instead of "controller"?  Otherwise this PR looks good

stuggi: done

dprince (approved):

PR #206, opened 2021-04-13
Allow baseImageVolumeName to be truly optional

abays:
Moves the initialization/generation of the `baseImageName` to earlier in the VMSet logic, and sets either the default or the CR spec's value as its value.  That value is then passed to the functions that need it.

### DIFF
--- a/controllers/openstackvmset_controller.go
+++ b/controllers/openstackvmset_controller.go
@@ -293,9 +293,14 @@ func (r *OpenStackVMSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
 	}
 
+	baseImageName := fmt.Sprintf("osp-vmset-baseimage-%s", instance.UID[0:4])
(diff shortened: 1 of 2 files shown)


controllers/openstackvmset_controller.go
```diff
@@ -435,7 +437,7 @@ func (r *OpenStackVMSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	// Create controller VM objects and finally set VMSet status in etcd
 	if instance.Status.BaseImageDVReady {
 		for _, ctl := range controllerDetails {
-			err = r.vmCreateInstance(instance, envVars, &ctl, osNetBindings)
+			err = r.vmCreateInstance(instance, envVars, &ctl, osNetBindings, baseImageName)
```
stuggi, line 440: we could add the baseimage name as a new parameter to the controllerDetails ospdirectorv1beta1.Host{} then we do not need to pass a new parameter to the vmCreateInstance method.

stuggi (approved):
/lgtm

PR #207, opened 2021-04-14
baseImageVolumeName README additions

abays:
Shows how to use `virtctl` to pre-load the RHEL base image


dprince (approved):

dprince:
Looks good to me. At some point we'll need to remove internal URL's and use real ones but this is great for now.

stuggi (commented):
/lgtm

PR #209, opened 2021-04-16
Image downloader Dockerfile to meet CPaaS requirements

mpryc:
Updated Dockerfile for the image_downloader container that
allows to translate from upstream to downstream using
dockerfile_to_osbs.sh from openstack-k8s-operators-ci.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,15 +12,10 @@ ARG REMOTE_SOURCE_DIR=/remote-source
 ARG REMOTE_SOURCE_SUBDIR=
 ARG DEST_ROOT=/dest-root
 
-ARG GO_BUILD_EXTRA_ARGS
+ARG GO_BUILD_EXTRA_ARGS=
 
 COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR
 WORKDIR ${REMOTE_SOURCE_DIR}/${REMOTE_SOURCE_SUBDIR}
-
-RUN go version
-RUN go env
-RUN if [ -f $CACHITO_ENV_FILE ] ; then echo $CACHITO_ENV_FILE; cat $CACHITO_ENV_FILE ; fi
-RUN if [ -f $CACHITO_ENV_FILE ] ; then source $CACHITO_ENV_FILE ; fi && go env
 
 RUN mkdir -p ${DEST_ROOT}/usr/local/bin/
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -52,7 +45,7 @@ LABEL com.redhat.component="osp-director-operator-container" \
 
 ENV USER_UID=1001 \
     OPERATOR_TEMPLATES=/usr/share/osp-director-operator/templates/ \
-    WATCH_NAMESPACE=openstack,openshift-machine-api,openshift-sriov-network-operator
```
yazug, line 55: is this intentional change for this effort? This change should likely be driven based on the operator change not a rework for layout/structure of the Dockerfile

mpryc: @yazug This is what I found in the midstream repo, but I can revert back to the upstream version of it.

yazug: use upstream version please.

Dockerfile.provision-ip-discovery-agent
```diff
@@ -1,15 +1,45 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
-WORKDIR /go/src/github.com/openstack-k8s-operators/osp-director-operator
-COPY . .
-RUN go mod vendor
-RUN BIN_PATH=build/_output/cmd ./containers/provision_ip_discovery_agent/build-provision-ip-discovery-agent.sh
```
yazug, line 5: hmm If I remember correctly this script does have some additional version injection stuff that wasn't done in here so not sure how to resolve the discrepancy and should probably get input from maintainer of that build script

Dockerfile.provision-ip-discovery-agent
```diff
@@ -1,15 +1,45 @@
...
+LABEL com.redhat.component="osp-director-provisioner-container" \
+      name="osp-director-provisioner-container" \
+      version="1.0"
+
+WORKDIR /
+
+# Install binary to WORKDIR
+COPY --from=builder ${DEST_ROOT}/${WHAT} .
+
+ENTRYPOINT ["/provision-ip-discovery-agent"]
```
yazug, line 45: hmm we flipped from CMD to ENTRYPOINT not sure if this is workable flip or improvement?

Dockerfile.provision-ip-discovery-agent
mpryc: I have used CPaaS version of Dockerfile rather then logic from here, I can add it, will check with @abays if it's required.

Dockerfile.provision-ip-discovery-agent
mpryc: It's almost the same, used ENTRYPOINT to be similar to operator Dockerfile as those two images should not be much different.

Dockerfile.provision-ip-discovery-agent
abays: The `VERSION_OVERRIDE` in the build script `containers/provision_ip_discovery_agent/build-provision-ip-discovery-agent.sh` did not end up ever being used, if I recall correctly.  Is this the version injection to which you were referring, @yazug?

yazug: yes, that is the part I was referring to, can we prune that out or the script out if this is sufficient?

abays: Yeah, just ignore that part of the bash script.  I borrowed it from somewhere else where the versioning stuff must have been needed, but then I didn't bother to pare it down for our version.

abays:
The two containers build and function as expected in my environment

abays (approved):
/lgtm

PR #210, opened 2021-04-20
Support deploy ComputeHCI

stuggi:
Adds support for Compute nodes also be Ceph OSP nodes.

Also splits the tripleo-deploy.sh to have -r and -p option for
the two phases:
-r use ephemeral heat to render the ansible playbooks
-p run ansible playbooks

With this we can control which part of the deployment process
to run.

Note: right now skip-tags opendev-validation-ceph gets added to
the deployment command if IsHCI is set on a baremetalset. Ceph
validations check verify if lvm2 package is installed. The check
verifies if the package is in any host of the ansible facts, which
includes the "undercloud". Since the lvm2 package is not installed
in the openstackclient container image this fails.

### DIFF
--- a/controllers/openstackcontrolplane_controller.go
+++ b/controllers/openstackcontrolplane_controller.go
@@ -143,7 +143,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			Role:                controlplane.Role,
 			HostCount:           controlplane.Count,
 			VIP:                 true,
-			AddToPredictableIPs: false,
+			AddToPredictableIPs: true,
 		}
 		ipset, op, err := common.OvercloudipsetCreateOrUpdate(r, instance, ipsetDetails)
 		if err != nil {
--- a/templates/openstackclient/bin/init.sh
+++ b/templates/openstackclient/bin/init.sh
@@ -16,6 +16,21 @@
 set -ex
 
 # if the pvc is an empty volume, copy the existing hosts file to it
-if [ ! -f /mnt/hosts ]; then
-  cp /etc/hosts /mnt
+if [ ! -f /mnt/etc/hosts ]; then
+  cp /etc/hosts /mnt/etc/
 fi
+
+mkdir -p /home/cloud-admin/tripleo-deploy/validations
+rm -rf /home/cloud-admin/tripleo-deploy/overcloud-ansible*
+
+# add cloud-admin ssh keys to EmptyDir Vol mount to /root/.ssh in openstackclient
+sudo mkdir -p /root/.ssh
+sudo cp /mnt/ssh-config/* /root/.ssh/
+sudo chmod 600 /root/.ssh/id_rsa
+sudo chown -R root: /root/.ssh
+
+# add cloud-admin ssh keys to /home/cloud-admin/.ssh in openstackclient
+mkdir -p /home/cloud-admin/.ssh
+cp /mnt/ssh-config/* /home/cloud-admin/.ssh/
+chmod 600 /home/cloud-admin/.ssh/id_rsa
+chown -R cloud-admin: /home/cloud-admin/.ssh
--- a/templates/openstackclient/bin/tripleo-deploy.sh
+++ b/templates/openstackclient/bin/tripleo-deploy.sh
@@ -2,16 +2,20 @@
 
 set -eux
 
-# disable running dhcp on all interfaces, setting disable_configure_safe_defaults in the interface template does not work
-sudo sed -i '/^set -eux/a disable_configure_safe_defaults=true' /usr/share/openstack-tripleo-heat-templates/network/scripts/run-os-net-config.sh
+render() {
+  # disable running dhcp on all interfaces, setting disable_configure_safe_defaults in the interface template does not work
+  sudo sed -i '/^set -eux/a disable_configure_safe_defaults=true' /usr/share/openstack-tripleo-heat-templates/network/scripts/run-os-net-config.sh
 
-mkdir -p ~/tripleo-deploy
-rm -rf ~/tripleo-deploy/overcloud-ansible*
-unset OS_CLOUD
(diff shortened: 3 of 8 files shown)


controllers/openstackcontrolplane_controller.go
```diff
@@ -143,7 +143,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			Role:                controlplane.Role,
 			HostCount:           controlplane.Count,
 			VIP:                 true,
-			AddToPredictableIPs: false,
+			AddToPredictableIPs: true,
```
abays, line 146: Oh, have we had this wrong all along?  Or is this necessary for these new changes somehow?

templates/openstackclient/bin/init.sh
```diff
@@ -17,5 +17,20 @@ set -ex
 
 # if the pvc is an empty volume, copy the existing hosts file to it
 if [ ! -f /mnt/hosts ]; then
-  cp /etc/hosts /mnt
+  cp /etc/hosts /mnt/hosts/
```
abays, line 20: Should this maybe be `/mnt/hosts`, i.e. leave off the trailing `/`?  Just checking

stuggi: no, I changed to mount the volume where we at the moment store the permanent /etc/hosts file from /mnt to /mnt/hosts, maybe /mnt/etc is a better choice

controllers/openstackcontrolplane_controller.go
stuggi: I realized that /etc/hosts entries from the overcloud deployment were wrong. instead of the overcloud VIPs we added the openstackclient ips ...

abays (commented):
Nice work @stuggi!  I'll defer to @dprince on the approval, but it seems good.

stuggi:
ceph was falling back to ctlplane network, identified the issue with isolated networks. converted to draft until tested

stuggi:
network isolation issue fixed in https://github.com/openstack-k8s-operators/osp-director-operator/pull/214

templates/openstackclient/bin/init.sh
stuggi: changed this to /mnt/etc to be more clear

templates/openstackclient/bin/tripleo-deploy.sh
```diff
@@ -2,21 +2,28 @@
...
+  sudo openstack tripleo deploy \
     --templates /usr/share/openstack-tripleo-heat-templates \
-    -r /usr/share/openstack-tripleo-heat-templates/roles_data.yaml \
+    -r ROLESFILE \
     -n /usr/share/openstack-tripleo-heat-templates/network_data.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/overcloud-resource-registry-puppet.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/environments/network-isolation.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/environments/deployed-server-environment.yaml \
     -e /usr/share/openstack-tripleo-heat-templates/environments/docker-ha.yaml \
+{{- if .EnableCeph }}
```
dprince, line 24: Would it be reasonable to have the user just supply this via the custom-config ConfigMap instead? My concern is that logic like this may not be scalable as we add features.

dprince (commented):
One inline comment. I'd also like to ponder the tripleo-deploy.sh changes a bit more as my incoming patch would pretty much remove this entirely. Do we even need an ansible runner script after its all said and done?

templates/openstackclient/bin/tripleo-deploy.sh
stuggi: yes we can do that, then we won't need the isHCI at all and required the user to pass all information from that ceph env file into the custom-config ConfigMap, or it might be a good idea to provide an interface a user can use to specify which ../environments/.. files provided by triplo should be added to the deploy script?

stuggi:
> One inline comment. I'd also like to ponder the tripleo-deploy.sh changes a bit more as my incoming patch would pretty much remove this entirely.

yes I know that this will completely change with what you work on right now. it was just a useful change to split the workflow whil I was working on this.

> Do we even need an ansible runner script after its all said and done?

hmm, not sure I get that question :)

templates/openstackclient/bin/tripleo-deploy.sh
dprince: Yes. I think the OSPlaybookGenerator might be a good place for that.

dprince:
> > Do we even need an ansible runner script after its all said and done?
> 
> hmm, not sure I get that question :)

What I meant was do we need a script to run ansible once we merge the "playbook generator" patch. I guess it might be useful but I didn't have one in there yet.

stuggi:
> > > Do we even need an ansible runner script after its all said and done?
> > 
> > 
> > hmm, not sure I get that question :)
> 
> What I meant was do we need a script to run ansible once we merge the "playbook generator" patch. I guess it might be useful but I didn't have one in there yet.

I think it is still useful, for debugging, or limit to run things only on specific overcloud nodes

controllers/openstackclient_controller.go
```diff
@@ -209,6 +209,16 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	osbmsList := &ospdirectorv1beta1.OpenStackBaremetalSetList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(instance.Namespace),
+	}
+
+	err = r.Client.List(context.TODO(), osbmsList, listOpts...)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
dprince, line 221: Are we actually using osbmsList anywhere?

stuggi: no, no longer, removed in latest update

PR #217, opened 2021-04-30
Use biosdev nic naming for overcloud nodes

stuggi:
Document usage of biosdev names for osvmset and osbmset and
introduce parameter to set ctlplane interface for vmset, which
defaults to enp2s0 and usually won't need to be changed.


stuggi:
should we also use biosdev names on the controller vms?

abays (approved):
/lgtm

PR #223, opened 2021-05-05
OSNet: Optimize the Osipset configmap writes

dprince:
Only write configmap if the reservation count equals the
count of the total osipsets

### DIFF
--- a/controllers/openstackipset_controller.go
+++ b/controllers/openstackipset_controller.go
@@ -212,6 +212,14 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	// Only write config maps if the processed reservations are equal to the number of OSIPSets for each network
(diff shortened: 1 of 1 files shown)


controllers/openstackipset_controller.go
```diff
@@ -212,6 +212,14 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
+	// Only write config maps if the processed reservations are equal to the number of OSIPSets for each network
+	// NOTE: don't put any reconcile logic below this except ConfigMap things as this will short circuit the reconciliation
+	for _, osnet := range overcloudNetList.Items {
+		if len(osnet.Status.Reservations) != len(overcloudIPList.Items) {
+			return ctrl.Result{}, nil
```
stuggi, line 219: we should add a log message on which network is the one, and reconcile after X seconds or we need to watch osnets?

dprince: I was going to let the last osnet write the file and be done with it. If we requeue it would cause all the ipsets to re-reconcile.

PR #225, opened 2021-05-06
Add param to debug the playbook job

dprince:


pkg/openstackplaybookgenerator/job.go
```diff
@@ -44,6 +44,11 @@ func PlaybookJob(cr *ospdirectorv1beta1.OpenStackPlaybookGenerator, configHash s
 	volumeMounts := GetVolumeMounts(cr)
 	volumes := GetVolumes(cr)
 
+	cmd := []string{"/bin/bash", "/home/cloud-admin/create-playbooks.sh"}
+	if cr.Spec.Debug {
+		cmd = []string{"/bin/sleep", "360d"}
```
stuggi, line 49: should we use sleep infinity?

abays: Same question from me, but otherwise looks good.

dprince: infinity it is

stuggi (approved):
/lgtm

PR #231, opened 2021-05-11
BaremetalSet status improvements

abays:
Adds additional CR status to `OpenStackBaremetalSet` and surfaces it in the CLI


config/crd/bases/osp-director.openstack.org_openstackbaremetalsets.yaml
```diff
@@ -50,6 +67,10 @@ spec:
                 description: BmhLabelSelector allows for a sub-selection of BaremetalHosts
                   based on arbitrary labels
                 type: object
+              count:
+                description: Replicas The number of baremetalhosts to attempt to aquire
```
stuggi, line 71: remove Replicas, see above

api/v1beta1/openstackbaremetalset_types.go
```diff
@@ -27,7 +27,7 @@ type HardwareReqType string
 type OpenStackBaremetalSetSpec struct {
 	// Replicas The number of baremetalhosts to attempt to aquire
```
stuggi, line 28: Replicas => Count ?

controllers/openstackbaremetalset_controller.go
```diff
@@ -223,26 +252,85 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
...
 	}
 
 	// Provision / deprovision requested replicas
-	if err := r.ensureBaremetalHosts(instance, provisionServer, sshSecret, ipset); err != nil {
+	if err := r.ensureBaremetalHosts(instance, provisionServer, sshSecret, passwordSecret, ipset); err != nil {
+		// As we know, this transient "object has been modified, please try again" error occurs from time
+		// to time in the reconcile loop.  Let's avoid counting that as an error for the purposes of
+		// status display
+		// FIXME?: Is there an "errors.IsWhatever()" func we can call for this type of error?
+		if !strings.Contains(err.Error(), "please apply your changes to the latest version and try again") {
```
stuggi, line 271: if we hit this and don't want to reconcile with an error, we might want to reconcile without logging an error and do a manual retry like:
   return ctrl.Result{RequeueAfter: time.Second * 1}, nil

Instead of hard code this string here, we can use the OptimisticLockErrorMsg const from
https://pkg.go.dev/k8s.io/apiserver/pkg/registry/generic/registry#pkg-constants

abays: Ah, thank you!  That seems to fit what I was looking for in the `FIXME`.

api/v1beta1/openstackbaremetalset_types.go
abays: Ack

config/crd/bases/osp-director.openstack.org_openstackbaremetalsets.yaml
abays: Ack

dprince:
Aside from the inline comments I'm good with this

stuggi (approved):
/lgtm

PR #234, opened 2021-05-12
rabbitmq: bump RABBITMQ_IO_THREAD_POOL_SIZE

dprince:
This bumps the RABBITMQ_IO_THREAD_POOL_SIZE to 128 to avoid
"Failed to create thread" aborts on some machines with a larger
number of cores.

### DIFF
--- a/pkg/openstackephemeralheat/rabbitmq.go
+++ b/pkg/openstackephemeralheat/rabbitmq.go
@@ -27,6 +27,11 @@ func RabbitmqPod(instance *ospdirectorv1beta1.OpenStackEphemeralHeat) *corev1.Po
 					Name:  "rabbitmq",
 					Image: instance.Spec.RabbitImageURL,
 					Env: []corev1.EnvVar{
+						// added to resolve "Failed to create thread" aborts
+						{
+							Name:  "RABBITMQ_IO_THREAD_POOL_SIZE",
+							Value: "128",
+						},
 						{
 							Name:  "KOLLA_CONFIG_STRATEGY",
 							Value: "COPY_ALWAYS",


pkg/openstackephemeralheat/rabbitmq.go
```diff
@@ -27,6 +27,11 @@ func RabbitmqPod(instance *ospdirectorv1beta1.OpenStackEphemeralHeat) *corev1.Po
 					Name:  "rabbitmq",
 					Image: instance.Spec.RabbitImageURL,
 					Env: []corev1.EnvVar{
+						// added to resolve "Failed to create thread" aborts
+						{
+							Name:  "RABBITMQ_IO_THREAD_POOL_SIZE",
+							Value: "128",
```
stuggi, line 33: in a follow up, we could do this more dynamic like in [1] based on the number of available cores on the worker node. Not sure what version we have, just wanted to reference this from [2] "That setting is no longer relevant on Erlang 21+ [3] and will be removed for 3.8.10."

[1] https://opendev.org/starlingx/upstream/commit/3209f7cdad1df7b9e36990c43a7bee9101d77b6d
[2] https://groups.google.com/g/rabbitmq-users/c/ddBWt8lq9to
[3] https://github.com/rabbitmq/rabbitmq-server/issues/2473

stuggi (commented):
/lgtm

pkg/openstackephemeralheat/rabbitmq.go
snecklifter:
```
bash-4.4$ erl -eval '{ok, Version} = file:read_file(filename:join([code:root_dir(), "releases", erlang:system_info(otp_release), "OTP_VERSION"])), io:fwrite(Version), halt().' -noshell
22.3.4.1
bash-4.4$ rabbitmqctl --version
3.8.3
```

dprince: Agreed. I filed an issue for us to track this here: https://github.com/openstack-k8s-operators/osp-director-operator/issues/235

PR #238, opened 2021-05-14
Allow mysql to read its config

dprince:
This resolves issues where mysqld wasn't reading the
provided config file


abays (approved):
/lgtm

dprince:
This works but breaks the Heat API pod. So Need to fix that before we merge :/

abays (approved):
/lgtm

Issue #239, opened 2021-05-14
Fix missing Image (heat stack create failures) in new playbook generator.


dprince:
Here is a snippet from the heat stack create failure:
2021-05-13 13:52:02Z [overcloud.ComputeServiceChain]: CREATE_FAILED  resources.ServiceChain: resources.ComputeServiceChain.Property error: resources[46].properties: Property ContainerOv
nMetadataImage not assigned
2021-05-13 13:52:02Z [overcloud]: CREATE_FAILED  Resource CREATE failed: resources.ServiceChain: resources.ComputeServiceChain.Property error: resources[46].properties: Property Contain
erOvnMetadataImage not assigned

PR #240, opened 2021-05-14
Fix 'Property Container*Image not assigned' errors

dprince:
This error somehow crept into the recent osplaybookgenerator workflow.
We were sending all the environments to the 'container image prepare'
command but only the container-image-prepare.yaml is really required.


dprince:
The issue https://github.com/openstack-k8s-operators/osp-director-operator/issues/239

stuggi (approved):
/lgtm

PR #241, opened 2021-05-17
Add git support for playbookgenerator

dprince:


controllers/openstackplaybookgenerator_controller.go
```diff
@@ -287,6 +287,11 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 		requeue, err := common.WaitOnJob(job, r.Client, r.Log)
 		r.Log.Info("Generating Playbooks...")
 		if err != nil {
+			// the job failed in error
+			deleteErr := r.Client.Delete(context.TODO(), heat)
+			if err != nil {
```
abays, line 292: Should this be `deleteErr` instead of `err`?

dprince: Yes. Just pushed an update to fix this.

templates/openstackplaybookgenerator/bin/create-playbooks.sh
```diff
@@ -132,5 +147,19 @@ ansible.write_default_ansible_cfg(
 
 EOF_PYTHON
 
-# copy to our persistent volume mount so it shows up in the 'openstackclient' pod
-cp -a /home/cloud-admin/ansible/* /var/cloud-admin/ansible/
+TMP_DIR=$(mktemp -d)
+git clone $GIT_URL $TMP_DIR
+pushd $TMP_DIR
```
bogdando, line 152: without popd, wouldn't this end up with the dirs stack growing up trashed with tempdirs?

api/v1beta1/openstackclient_types.go
```diff
@@ -27,6 +27,8 @@ type OpenStackClientSpec struct {
 	ImageURL string `json:"imageURL"`
 	// name of secret holding the stack-admin ssh keys
 	DeploymentSSHSecret string `json:"deploymentSSHSecret"`
+	// GitSecret used to pull playbooks into the openstackclient pod
```
bogdando, line 30: nit: create-playbooks.sh pushes by git_url and git secret as well

config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/openstack-k8s-operators/osp-director-operator
-  newTag: 0.0.1
+￼ newTag: 0.0.1
```
stuggi, line 8: is this to change a bad char, or introduce it :)

templates/openstackplaybookgenerator/bin/create-playbooks.sh
dprince: The entire pod is ephemeral and is destroyed (we are using emptydir for local storage with these pods). So I think its okay

config/manager/kustomization.yaml
dprince: yes, it is a bad char. I think its gone in the latest commit.

abays:
This seems to be ready for merge, or did you have other changes you wanted to add @dprince?

dprince:
I just pushed a few README and sample config changes but nothing major in that last commit. Functionally what is here works great with the dev-tools patch.

There is one reconciliation bug I am tracking down. The Ephemeral Heat is getting created and immediately deleted (which looks bad if you notice it but doesn't cause a functional issue). Other than that this can land and I think is in good shape

stuggi (approved):
/lgtm

PR #246, opened 2021-05-23
IP handling during scaling

stuggi:
Scaling down has the following workflow:
1) annotate VM or BMH which should be removed
2) decrease the resource count of OSBMset/OSCtlplane

Right now the IP reservation entry won't be set as deleted.
As a result, we'll still have the deleted hosts in the
rendered predictible IPs list of the tripleo environment
file.

This change introduce the following new behavior:
* annotate VM or BMH with a unified delete annotation
  osp-director.openstack.org/delete-host=true
* the annotation status of the VM/BMH gets reflected in the
  resource status via "annotatedForDeletion": false/true
* decrease the resource count of OSBMset/OSCtlplane to
  remove the annotated VM/BMH
* when node gets deleted it is deleted from
  - the resource CR status (OSVMSet/OSBMset)
  - the node gets removed from the ipset status
  - the osnet status entry get set as deleted
  -> as a result the rendered tripleo IPs list for the role
     will have a deleted entry
* right now the IPreservation is only marked as deleted,
  if a new node gets scaled in it will get the same hostname
  and IP address.

The osnet status is now a map[string]OpenStackNetRoleStatus,
where the OpenStackNetRoleStatus
- has the flag if all entries from that role should be added
  to the tripleo predictible IPs list (AddToPredictableIPs)
- list of IPReservations

With this structur, if the corresponding role CR gets deleted
the IP reservations can all be deleted at once via the added
finalizer in the OSIPset controller.

Additional changes:
- use common LabelSelectors for controllers
- simplifies the IPSet status.network status to not contain
  AttachConfiguration specifics


README.md
```diff
@@ -456,3 +455,120 @@ ansible -i inventory.yaml overcloud -a "sudo dnf -y install python3 lvm2"
...
+```
+
+The annotation status is being reflected in the OSBaremetalset/OSVMset using the `annotatedForDeletion` parameter:
+
+```bash
+oc get osbms computehci -o json | jq .status
+{
+  "baremetalHosts": {
+    "computehci-0": {
+      "annotatedForDeletion": true,
```
stuggi, line 493: should we reflect this in the new status summary, either annotated for deletion count, or list of names?

abays: Hmm.  If the list doesn't make for too much clutter, we could do that.  If it is, then perhaps a count would work.

README.md
```diff
@@ -456,3 +455,120 @@ ansible -i inventory.yaml overcloud -a "sudo dnf -y install python3 lvm2"
...
+```
+
+This results in the following behavior
+* the IP is not free for use for another role
+* if a new node gets scaled into the same role it will reuse the hostnames starting with lowest id suffix (if there are multiple) _and_ corresponding IP reservation
+* if the OSBaremetalset or OSVMset resource gets deleted, all IP reservations for the role get deleted and are free to be used by other nodes
+
+## Cleanup OpenStack resources
+
+If a compute node got removed,
```
abays, line 559: Looks like you meant to write something more here

README.md
```diff
@@ -456,3 +455,120 @@ ansible -i inventory.yaml overcloud -a "sudo dnf -y install python3 lvm2"
...
+      "vip": false
+    }
+  ]
+}
+```
+
+This results in the following behavior
+* the IP is not free for use for another role
+* if a new node gets scaled into the same role it will reuse the hostnames starting with lowest id suffix (if there are multiple) _and_ corresponding IP reservation
+* if the OSBaremetalset or OSVMset resource gets deleted, all IP reservations for the role get deleted and are free to be used by other nodes
```
abays, line 555: Is this going to cause problems on the OOO side?  Just asking because my understanding was that we always needed to maintain past reservations, but maybe wiping a whole compute role or control plane role mitigates that concern?

controllers/openstackbaremetalset_controller.go
```diff
@@ -238,19 +237,127 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	// get list of all BMH
+	baremetalHostsList, err := common.GetBmhHosts(r, "openshift-machine-api", map[string]string{
+		common.OwnerControllerNameLabelSelector: baremetalset.AppLabel,
+	})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// check for deletion marked VMs
```
abays, line 297: Nit: I think you meant BMHs

controllers/openstackbaremetalset_controller.go
```diff
@@ -238,19 +237,127 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
 		instance.Status.BaremetalHosts = map[string]ospdirectorv1beta1.OpenStackBaremetalHostStatus{}
 	}
 
+	//
+	// create hostnames for the requested number of systems
+	//
+	//   create new host allocations count
+	//
```
abays, line 244: I wonder if there's a way to combine this add/delete logic with what we have in `ensureBaremetalHosts` func.  It seems like they do similar things in regards to the IPSet.  That might be tricky though, or perhaps I'm not understanding something.

controllers/openstackclient_controller.go
```diff
@@ -94,17 +94,63 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		}
+
+		err = common.CreateOrGetHostname(instance, &hostnameDetails)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	} else {
+		for hostname := range hostnameRef {
+
+			hostnameDetails = common.Hostname{
```
abays, line 126: It looks like `hostnameDetails` will always end up being the last element of `hostnameRef`.  Just want to make sure that's what you intended

README.md
stuggi: no, this does not cause issues. for existing roles we only need to make sure we have a persistent list of max predictable IPs list entries, like:
~~~
      ComputeHCIIPs:
        internal_api:
        - 172.16.2.9
        - 172.16.2.10
        - 172.16.2.11
        - 172.16.2.12
~~~

When compute-2 gets deleted we must keep an entry, not necessarily the IP reservation, to have:
~~~
      ComputeHCIIPs:
        internal_api:
        - 172.16.2.9
        - 172.16.2.10
        - deleted
        - 172.16.2.12
~~~

controllers/openstackclient_controller.go
stuggi: for openstackclient we always only have 0 or 1 hostnames.

README.md
stuggi: yes that's the only concern that the list could get to long

README.md
stuggi: oh right :)

controllers/openstackbaremetalset_controller.go
stuggi: yep, thanks

controllers/openstackbaremetalset_controller.go
stuggi: we probably could. we just need to make sure we do the hostname reservation/creation before we to the IP reservation to be able to pass the hostname list. we'd then also need to move the ipset part to ensureBaremetalHosts

controllers/openstackbaremetalset_controller.go
stuggi: done

README.md
stuggi: done

controllers/openstackipset_controller.go
```diff
@@ -246,6 +245,199 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		allActiveIPSetHosts = append(allActiveIPSetHosts, hostname)
+	}
+
+	// mark entry in OSNet status as deleted
+	for _, netName := range instance.Spec.Networks {
+		network := &ospdirectorv1beta1.OpenStackNet{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: netName, Namespace: instance.Namespace}, network)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				r.Log.Info(fmt.Sprintf("OpenStackNet named %s not found!", netName))
```
abays, line 274: Do we maybe want to `continue` here?  Just wondering if you meant to do something in addition to printing the error message.  It will just return the err, but that could be what you want.

controllers/openstackipset_controller.go
```diff
@@ -246,6 +245,199 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		hostNetworkIPs := map[string]string{}
+
+		// iterate over the requested Networks
+		for _, netName := range instance.Spec.Networks {
+
+			network := &ospdirectorv1beta1.OpenStackNet{}
+			err := r.Client.Get(context.TODO(), types.NamespacedName{Name: netName, Namespace: instance.Namespace}, network)
+			if err != nil {
+				if k8s_errors.IsNotFound(err) {
+					r.Log.Info(fmt.Sprintf("OpenStackNet named %s not found!", netName))
```
abays, line 318: Same question here as line 274.  Or perhaps it would be better to catch the `NotFound` error in the caller and return a requeue request at X seconds?

controllers/openstackipset_controller.go
```diff
@@ -246,6 +245,199 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+func (r *OpenStackIPSetReconciler) cleanupOSNetStatus(instance *ospdirectorv1beta1.OpenStackIPSet) error {
+
+	// mark entry in OSNet status as deleted
+	for _, netName := range instance.Spec.Networks {
+		network := &ospdirectorv1beta1.OpenStackNet{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: netName, Namespace: instance.Namespace}, network)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				r.Log.Info(fmt.Sprintf("OpenStackNet named %s not found!", netName))
```
abays, line 425: Same question here regarding the `continue` consideration from above

controllers/openstackipset_controller.go
```diff
@@ -246,6 +245,199 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+			if reservationIP == "" {
+				r.Log.Info(fmt.Sprintf("Create new reservation for host %s in network %s", hostname, netName))
+
+				// No reservation found, so create a new one
+				ip, reservation, err := common.AssignIP(common.AssignIPDetails{
+					IPnet:           *cidr,
+					RangeStart:      start,
+					RangeEnd:        end,
+					RoleReservelist: network.Status.RoleReservations[instance.Spec.RoleName].Reservations,
```
abays, line 371: Is it possible that `network.Status.RoleReservations[instance.Spec.RoleName]` could point to a `nil` or zero-value entry?  If so, maybe that could cause a problem when it then tries to access non-existent `.Reservations`.

controllers/openstackipset_controller.go
stuggi: yes, just wanted to record the missing network and continue as we don't have to sync deleted ips in that case. I'll add continue

controllers/openstackipset_controller.go
stuggi: indeed thats better

controllers/openstackipset_controller.go
stuggi: yes, agreed

controllers/openstackipset_controller.go
stuggi: change to initialize them if it does not exist

controllers/openstackbaremetalset_controller.go
```diff
@@ -436,10 +529,18 @@ func (r *OpenStackBaremetalSetReconciler) ensureBaremetalHosts(instance *ospdire
 		for i := 0; i < oldBmhsToRemoveCount; i++ {
 			// First choose BaremetalHosts to remove from the prepared list of BaremetalHosts
 			// that have the "osp-director.openstack.org/baremetalset-delete-baremetalhost=yes" annotation
```
abays, line 531: I must have missed this last review, but we should probably change the annotation comment here so it's not referring to the old annotation (just to avoid potentially confusing people)

controllers/openstackbaremetalset_controller.go
```diff
@@ -436,10 +529,18 @@ func (r *OpenStackBaremetalSetReconciler) ensureBaremetalHosts(instance *ospdire
...
 
 			if len(removalAnnotatedBaremetalHosts) > 0 {
-				err := r.baremetalHostDeprovision(instance, removalAnnotatedBaremetalHosts[0])
+				// get OSBms status corresponding to bmh
+				bmhStatus := ospdirectorv1beta1.OpenStackBaremetalHostStatus{}
+				for _, bmhStatus = range instance.Status.BaremetalHosts {
+					if bmhStatus.HostRef == removalAnnotatedBaremetalHosts[0] {
+						break
+					}
+				}
```
abays, line 541: Could use your new `getBmhHostRefStatus` function defined on lines 659-669 instead?

controllers/openstackbaremetalset_controller.go
stuggi: sure, good catch. found it also in another place

controllers/openstackbaremetalset_controller.go
stuggi: ack, will do

controllers/openstackbaremetalset_controller.go
stuggi: done

controllers/openstackbaremetalset_controller.go
stuggi: done

abays:
Nice work!  Seems to be ready for merge, but let's wait for @dprince to review.

dprince (approved):
/ltgm

dprince (approved):

PR #249, opened 2021-05-26
Fixes to playbookgenerator controller reconcile

dprince:
This adds some appropriate requeue's, and quiets down
some logging that were occuring when this controller
reconciled a playbookgenerator.


abays (approved):
This looks good to me.  I actually made a few of these changes myself in my upcoming patch.

PR #252, opened 2021-05-27
OpenStackPlaybookGenerator gui status

abays:
Adds conditions to `OpenStackPlaybookGenerator` status, so that the GUI can reflect the actual state of the CR.  Also adds a couple extra states for the OSPG to further refine the precision.


stuggi (commented):
lgtm , but want to get Dan's input as well

dprince (approved):

PR #255, opened 2021-06-01
For now set OVNMacAddressNetwork and OVNMacAddressPort to None

stuggi:
"Add support in tripleo for distributed vlan traffic in ovn"
( https://bugs.launchpad.net/tripleo/+bug/1881593 ) was backported
to 16.2, which creates an OVNMacAddressNetwork and OVNMacAddressPort
per compute to configure the ovn chassis.
Right now we don't support MAC address management via the operator.
For now set OVNMacAddressNetwork and OVNMacAddressPort to None, which
should not affect the default geneve NeutronNetworkType. If VLAN is
used, FIP traffic does not pass to a VLAN tenant network with ML2/OVN
and DVR. DVR is enabled by default. If you need VLAN tenant networks
with OVN, you can disable DVR. To disable DVR, inlcude the following
lines in an environment file:

```yaml
parameter_defaults:
  NeutronEnableDVR: false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,16 @@ A "Tarball Config Map" can be used to provide (binary) tarballs which are extrac
 
 -[Net-Config files](https://github.com/openstack-k8s-operators/osp-director-dev-tools/tree/master/ansible/files/osp/net_config).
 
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

PR #259, opened 2021-06-07
OpenStackNet CLI and GUI status improvements

abays:


controllers/openstacknet_controller.go
```diff
@@ -197,11 +220,97 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 
-	return ctrl.Result{}, nil
+	// Count IP assignments and update status
+	totalIps := 0
+	activeIps := 0
+
+	for _, roleIps := range instance.Status.RoleReservations {
+		for _, roleIP := range roleIps.Reservations {
+			totalIps++
+			if !roleIP.Deleted {
```
stuggi, line 284: this could be confusing. the behavior right now is
- on a osvmset or osbmset delete, like role delete all IP reservations get returned to the pool
- on single system removal from an CR, the IP reservation is marked as deleted, but can not be consumed by another CR/role, so it is not completely deleted. if we have e.g. a role which used 50 IPs, then remove 40 nodes from that group, this will still have the 40 as deleted, but shown as only 10 be consumed, right?

Maybe we should just show the len of roleIps.Reservations and in a follow up discuss if we should return all deleted flagged IP reservations to be available for any role?

controllers/openstacknet_controller.go
```diff
@@ -197,11 +220,97 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	return ctrl.Result{}, err
+}
+
+func (r *OpenStackNetReconciler) setStatus(instance *ospdirectorv1beta1.OpenStackNet, oldStatus *ospdirectorv1beta1.OpenStackNetStatus, msg string) error {
+
+	if msg != "" {
+		r.Log.Info(msg)
+	}
+
+	if oldStatus.CurrentState != instance.Status.CurrentState || oldStatus.ActiveIPCount != instance.Status.ActiveIPCount || oldStatus.ReservedIPCount != instance.Status.ReservedIPCount {
```
stuggi, line 304: should we use DeepEqual instead of != to be in sync with osbms controller?

controllers/openstacknet_controller.go
```diff
@@ -197,11 +220,97 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		// have a status block we can examine), so now examine the status of the NNCP
+
+		//nncp := &nmstatev1alpha1.NodeNetworkConfigurationPolicy{}
+		nncpList := &nmstatev1alpha1.NodeNetworkConfigurationPolicyList{}
+
+		// Can't use "r.Client.Get" here, as the NNCP is cluster-scoped and does not have a namespace.  We require the
+		// NNCP resource struct object in order to examine its status, but the NNCP was created earlier via a server-side
+		// apply.  Thus we don't already have the struct object in memory, and must acquire it by asking for a list of
+		// NNCPs filtered by our label selector (which should return just the one resource we seek)
+		// err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ncp.Name, Namespace: instance.Namespace}, nncp)
```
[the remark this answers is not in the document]
abays: Sure, we can remove the commented code in this area (and above).  I meant it as a means of example to go along with the explanation, but it doesn't need to be there

controllers/openstacknet_controller.go
abays: We can certainly just show the sum of the length of all `roleIps.Reservations` for now and discuss further changes for later.  But yes, you're right.  In that scenario, if you had 50 and deleted 40, it would say 50 reserved and 10 active.

controllers/openstacknet_controller.go
abays: Sounds good.  Will change that.

stuggi (commented):
/lgtm

PR #265, opened 2021-06-09
Reduce osctlplane pod watch scope

stuggi:
The osctlplane watched all pods in the openstack namespace, which
can lead in immense reconcile if playbook generator recreates its
pods again and again. This limits to reconcile the osctlplane to
pods from vmset and openstackclient having the corrsponding AppLabel.


abays (approved):
/lgtm

PR #267, opened 2021-06-09
tripleo-deploy.sh init() git fix

dprince:
The init() function can fail to initialize the latest tag
if the the most recent branch is '-> HEAD'. This updates
the script to exclude that from the listing when detecting
the latest branch.


abays (approved):
/lgtm

PR #268, opened 2021-06-14
Reduce status updates

stuggi:
This introduce checkes when updating the status if the status
really changed from what is already there to prevent unnecessary
syncs to etcd.


controllers/openstackbaremetalset_controller.go
```diff
@@ -389,34 +405,53 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
...
+	if !reflect.DeepEqual(currentState, actualState) {
+		r.Log.Info(fmt.Sprintf("%s - diff %s", actualState.Reason, diff.ObjectReflectDiff(currentState, actualState)))
+		instance.Status.ProvisioningStatus = actualState
+
+		if err := r.Client.Status().Update(context.TODO(), instance); err != nil {
+			r.Log.Error(err, "Failed to update CR status %v")
+			return err
+		}
+	} else {
+		r.Log.Info(actualState.Reason)
```
abays, line 449: I think `actualState.reason` could be the empty string here sometimes.  i.e. if the new line 422 sets the `State` but does not set `Reason`.  If, on the other hand, there's an old `Reason` from a previous provisioning status, we should probably be sure to always set `State` and `Reason` whenever we change one or the other.

controllers/openstackvmset_controller.go
```diff
@@ -1041,3 +1099,27 @@ func (r *OpenStackVMSetReconciler) getDeletedVMOSPHostnames(instance *ospdirecto
...
+
+		instance.Status.Conditions = ospdirectorv1beta1.ConditionList{}
+		instance.Status.Conditions.Set(ospdirectorv1beta1.ConditionType(actualState.State), corev1.ConditionTrue, ospdirectorv1beta1.ConditionReason(actualState.Reason), actualState.Reason)
+
+		if err := r.Client.Status().Update(context.TODO(), instance); err != nil {
+			r.Log.Error(err, "Failed to update CR status %v")
+			return err
+		}
+	} else {
+		r.Log.Info(actualState.Reason)
```
abays, line 1121: Same thing here regarding empty string.  For instance, if the reconcile loop keeps hitting the new 663 line, `Reason` is just set to the empty string.  If we just check for this, we can prevent printing an empty `Info` log.

controllers/openstackbaremetalset_controller.go
stuggi: Ack, I can set the reason for that. I just used the same for those 2 states (provisioning/deprovisioning) as we have it right now https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstackbaremetalset_controller.go#L406-L409

stuggi: should we set a reason for those 2 states or check if reason is empty and don't log it.

abays: Whatever suits you and @dprince is fine with me.  If we want to add a reason to go along with the state always, we can certainly do that.  If not, yeah, let's check for the empty string and not log in such a case.

stuggi: I have pushed a commit to not log empty actualState.Reason messages. for vmset and bms

controllers/openstackvmset_controller.go
stuggi: I have pushed a commit to not log empty actualState.Reason messages. for vmset and bms

controllers/openstackbaremetalset_controller.go
stuggi: this is not good. during a deployment I see it still shows waiting on ipset as it is actually provisioning. I'll set a provisioning message

stuggi: now added status message for provisioning/deprovisioning and ignore other empty reasons

dprince (approved):

PR #271, opened 2021-06-15
Verify node status before run playbook job

stuggi:
This introduce a condtion to wait for all osbms and osvmset to
be in provisioned state before running the playbook rendering
job. With this we can be sure all resources are up before we
generate the playbooks or the user can run them.


abays (commented):
Seems ready to go to me

dprince:
@stuggi Do we need to add permissions to 'list' the vmset and baremetalset resources on the osplaybookgenerator controller?

stuggi:
> @stuggi Do we need to add permissions to 'list' the vmset and baremetalset resources on the osplaybookgenerator controller?
yes, done

abays (approved):
/lgtm

PR #273, opened 2021-06-15
Remove rhos-release and document rhsm usage

stuggi:
This removes the installation of rhos-release package vi cloud-init
and documents the procedure how to register the overcloud nodes
via ansible to RHSM/Satellite.

Also moves the checkout of the accepted rendered playbooks to the
tripleo-deploy.sh -a action that the updated inventory.yaml file
is available to be used to register the overcloud nodes before
running the playbooks with -p.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/146

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,13 +342,51 @@ The osplaybookgenerator created above will automatically generate playbooks any 
 
 9) Login to the 'openstackclient' pod and deploy the OSP software via the rendered ansible playbooks. At this point all baremetal and virtualmachine resources have been provisioned within the OCP cluster.
 
+The `tripleo-deploy.sh` script supports three actions:
+* `-d` - show the `git diff` of the playbooks to the previous accepted
+* `-a` - accept the new available rendered playbooks and tag them as `latest`
+* `-p` - run the ansible driven OpenStack deployment
+
+a) check for new version of renered playbooks and accept them
+
+```bash
+oc rsh openstackclient
+bash
+cd /home/cloud-admin
+
+# (optional) show the `git diff` of the playbooks to the previous accepted
+./tripleo-deploy.sh -d
+
+# accept the new available rendered playbooks (if available) and tag them as `latest`
+./tripleo-deploy.sh -a
+```
+
+b) register the overcloud systems to required channels
+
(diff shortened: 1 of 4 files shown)


templates/baremetalset/cloudinit/userdata
```diff
@@ -12,6 +12,4 @@ chpasswd:
     root:{{ .NodeRootPassword }}
   expire: False
 {{- end }}
-runcmd:
-  - dnf install -y http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
-  - rhos-release 16.2
\ No newline at end of file
+runcmd:
```
abays, line 15: Is it okay to have a key with no value here?

stuggi: runcmd is a list of commands, so I expect it to be ok to have an empty list, but we can remove it since we don't run a command with this

README.md
```diff
@@ -342,13 +342,51 @@ The osplaybookgenerator created above will automatically generate playbooks any
...
+
+```bash
+oc rsh openstackclient
+bash
+cd /home/cloud-admin
+
+# (optional) show the `git diff` of the playbooks to the previous accepted
+./tripleo-deploy.sh -d
+
+# (optional) accept the new available rendered playbooks (if available) and tag them as `latest`
```
dprince, line 360: this actually isn't optional. If "new" playbooks have been generated and pushed to the repo you have to run with -a to accept them.

templates/openstackclient/bin/tripleo-deploy.sh
```diff
@@ -49,12 +49,6 @@ play() {
     echo "Then re-run '$0 -p' to run the new playbooks."
     exit
   fi
-  if git branch | grep " tripleo_deploy_working$" > /dev/null; then
-    git checkout tripleo_deploy_working >/dev/null
-    git reset --hard refs/tags/latest
-  else
-    git checkout -b tripleo_deploy_working latest
-  fi
```
dprince, line 57: I had this code here so that (on first execution) running tripleo-deploy.sh -p would run correctly without error. Also, my thought was the actual "checkout" belonged to -p since it would be actually using the local checkout anyway. So let me ask this: does tripleo-deploy.sh -p run cleanly (on a fresh git repo/install) now?

README.md
stuggi: correct, was wrong c/p

templates/openstackclient/bin/tripleo-deploy.sh
stuggi: right, had to add a check if /home/cloud-admin/playbooks/tripleo-ansible exists

PR #280, opened 2021-06-22
Change nncp to be network independent

stuggi:
At the moment we create for each network a nncp and a
network-attachment-definition corrensponding to the network name.
This results in duplicate nncp for the same bridge if more then
one network use the same bridge which can result in config race.

This patch changes the nncp creation to create the nncp for the
bridge instead of the network. With this we only create a single
nncp for required bridges.

Notes:
- this change only creates/updates the nncp for the desired
  config in the osnet cr. It does not verify if the config is
  different from what the current nncp might be. In case two
  networks use the same bridge name, but different interfaces,
  resulting from a typo/missconfig. The nncp gets updated and
  might result in an update loop.
- since the osnet controller waits for the nncp to be created
  before it creates the network-attachment-definition, the check
  for the nncp in the osvmset/osclient controller were not
  required and are now removed.
- at the moment there is no finalizer no finalizer on the nncp.
  It is not an issue right now as we do not cleanup the nncp on
  osnet cr delete. When we add nncp cleanup we need to make sure
  to add finalizers on the nncp for each network that the nncp
  only gets deleted when all networks are gone.

### DIFF
--- a/controllers/openstacknet_controller.go
+++ b/controllers/openstacknet_controller.go
@@ -160,20 +160,10 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
 			return ctrl.Result{}, err
 		}
 	} else {
-		// generate NodeNetworkConfigurationPolicy
-		ncp := common.NetworkConfigurationPolicy{
-			Name:                           instance.Name,
-			Labels:                         common.GetLabels(instance, openstacknet.AppLabel, map[string]string{}),
-			NodeNetworkConfigurationPolicy: instance.Spec.AttachConfiguration.NodeNetworkConfigurationPolicy,
-		}
-		err = common.CreateOrUpdateNetworkConfigurationPolicy(r, instance, instance.Kind, &ncp)
-		if err != nil {
-			instance.Status.CurrentState = ospdirectorv1beta1.NetError
-			_ = r.setStatus(instance, oldStatus, err.Error())
-			return ctrl.Result{}, err
-		}
 
-		// create NetworkAttachmentDefinition
+		//
+		// get the desired state
+		//
 		// the desired state is raw json, marshal -> unmarshal to parse it
 		desiredStateByte, err := json.Marshal(instance.Spec.AttachConfiguration.NodeNetworkConfigurationPolicy)
 		if err != nil {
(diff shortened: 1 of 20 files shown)


controllers/openstacknet_controller.go
```diff
@@ -192,7 +182,39 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			common.OwnerControllerNameLabelSelector: openstacknet.AppLabel,
+		}
+
+		// generate NodeNetworkConfigurationPolicy for the bridge
+		ncp := common.NetworkConfigurationPolicy{
+			Name:                           bridgeName.String(),
+			Labels:                         labelSelector,
+			NodeNetworkConfigurationPolicy: instance.Spec.AttachConfiguration.NodeNetworkConfigurationPolicy,
+		}
+		err = common.CreateOrUpdateNetworkConfigurationPolicy(r, instance, instance.Kind, &ncp)
```
abays, line 200: This is where your comment regarding thrashing could occur, right?  If two or more `OsNet`s use the same bridge but attach a different interface in the `desiredState`?

stuggi: yes correct. if different `desiredState` configs get used for networks, but have the same `bridgeName` this would end up in an nncp update loop. An idea could be to check for existing nncp and skip updating the current running nncp config and just log.

controllers/openstacknet_controller.go
```diff
@@ -192,7 +182,39 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		if err != nil {
+			instance.Status.CurrentState = ospdirectorv1beta1.NetError
+			_ = r.setStatus(instance, oldStatus, err.Error())
+			return ctrl.Result{}, err
+		}
+
+		// TODO: add finalizer to the NCP to prevent the bridge from being deleted if a network gets deleted
```
abays, line 207: Could another option be to have this `OSNet` finalizer get all other `OSNet`s in the namespace and then only delete the `NNCP` if no other `OSNet`s are using the bridge?

abays (commented):
I'm good with this.  We can address the finalizer issue in a separate PR.

controllers/openstacknet_controller.go
stuggi: yes, thats another possible implementation

controllers/openstacknet_controller.go
dprince: Could we create an up front validation on OsNets to guard against this case (same bridge, but different interfaces)? Like a webhook?

dprince (approved):

PR #285, opened 2021-06-23
fix vip creation

stuggi:
At the moment we run over the virtualMachineRoles and create for
the network list of the role vips for the network. If there are
multiple roles with different network list, last virtualMachineRoles
network list updates the ipset and VIPs won't get created.

This changes to create a uniq list of all networks for all
virtualMachineRoles and creates VIPs for this uniq list.

### DIFF
--- a/controllers/openstackcontrolplane_controller.go
+++ b/controllers/openstackcontrolplane_controller.go
@@ -196,41 +196,42 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 	currentStatus := instance.Status.DeepCopy()
 	vmSets := []*ospdirectorv1beta1.OpenStackVMSet{}
 
+	// create VIP for all networks
+	uniqNetworksList := r.createUniqNetworkList(instance)
+	ipsetDetails := common.IPSet{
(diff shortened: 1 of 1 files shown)


controllers/openstackcontrolplane_controller.go
```diff
@@ -480,3 +481,21 @@ func (r *OpenStackControlPlaneReconciler) setProvisioningStatus(instance *ospdir
...
+
+func (r *OpenStackControlPlaneReconciler) createUniqNetworkList(instance *ospdirectorv1beta1.OpenStackControlPlane) []string {
+
+	// create uniq list networls of all VirtualMachineRoles
+	networkList := make(map[string]int)
+	for _, vmRole := range instance.Spec.VirtualMachineRoles {
+		for _, network := range vmRole.Networks {
+			networkList[network] = 1
+		}
+	}
```
abays, line 493: Ah, you use a map here to avoid duplicates and then turn it into a list below.  Clever!

abays (approved):
Seems ready to me.  Just needs a linting fix

abays (approved):
/lgtm

PR #288, opened 2021-06-24
Generate TripleO passwords as ConfigMap

dprince:
This patch adds support for generating TripleO passwords as
a ConfigMap.

### DIFF
--- a/controllers/openstackcontrolplane_controller.go
+++ b/controllers/openstackcontrolplane_controller.go
@@ -86,7 +86,7 @@ func (r *OpenStackControlPlaneReconciler) GetScheme() *runtime.Scheme {
 func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	_ = r.Log.WithValues("controlplane", req.NamespacedName)
 
-	// Fetch the controller VM instance
+	// Fetch the controlplane instance
 	instance := &ospdirectorv1beta1.OpenStackControlPlane{}
 	err := r.Client.Get(context.TODO(), req.NamespacedName, instance)
 	if err != nil {
@@ -103,7 +103,32 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 	// Used in comparisons below to determine whether a status update is actually needed
 	newProvStatus := ospdirectorv1beta1.OpenStackControlPlaneProvisioningStatus{}
 
+	// ConfigMap - containing Tripleo Passwords
 	envVars := make(map[string]common.EnvSetter)
+	cmLabels := common.GetLabels(instance, controlplane.AppLabel, map[string]string{})
+
+	templateParameters := make(map[string]interface{})
+	templateParameters["TripleoPasswords"] = common.GeneratePasswords()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	cm := []common.Template{
+		{
+			Name:           "tripleo-passwords",
+			Namespace:      instance.Namespace,
+			Type:           common.TemplateTypeConfig,
+			InstanceType:   instance.Kind,
+			AdditionalData: map[string]string{},
+			Labels:         cmLabels,
+			ConfigOptions:  templateParameters,
+		},
+	}
+
+	err = common.EnsureConfigMaps(r, instance, cm, &envVars)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 
 	// Secret - used for deployment to ssh into the overcloud nodes,
 	//          gets added to the controller VMs cloud-admin user using cloud-init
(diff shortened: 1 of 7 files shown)


controllers/openstackcontrolplane_controller.go
```diff
@@ -103,7 +103,32 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
...
+
+	templateParameters := make(map[string]interface{})
+	templateParameters["TripleoPasswords"] = common.GeneratePasswords()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	cm := []common.Template{
+		{
+			Name:           "tripleo-passwords",
```
stuggi, line 118: is there a reason why we create tripleo-passwords as a CM instead of a secret?

controllers/openstackcontrolplane_controller.go
```diff
@@ -427,6 +452,7 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) err
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&ospdirectorv1beta1.OpenStackControlPlane{}).
 		Owns(&corev1.Secret{}).
+		Owns(&corev1.ConfigMap{}).
```
abays, line 455: If we do use `CM`, I think we need to add the kubebuilder annotation for the permissions.  Also, should we use a filter function here so that not every config map in the namespace wakes the controller on a change?

controllers/openstackcontrolplane_controller.go
dprince: You beat me to this. I'm going to move it to a Secret. This is quicker to just "throw in" here as we have the template stuff working with CMs. (I don't think we have equivalent functionality for Secrets).

dprince:
I have not yet fully verified all the new password formats are working. Ceph, MigrationKeys, Keystone, and Designate all required custom functions which we want to verify are working correctly.

go.mod
```diff
@@ -21,9 +22,10 @@ require (
 	github.com/tidwall/gjson v1.6.1
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
-	golang.org/x/tools v0.1.3 // indirect
+	golang.org/x/tools v0.1.4 // indirect
 	google.golang.org/grpc v1.31.0 // indirect
 	google.golang.org/grpc/examples v0.0.0-20210512000516-62adda2ece5e // indirect
+	gopkg.in/yaml.v2 v2.3.0 // indirect
```
dprince, line 28: yaml can go away I think (go mod tidy)

controllers/openstackcontrolplane_controller.go
stuggi: it should just work if you use common.EnsureSecrets [1] instead of common.EnsureConfigMaps in L128

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/secret.go#L219

dprince: sounds good. I didn't know we had that (or forgot)

abays (commented):
+1 from me

controllers/openstackcontrolplane_controller.go
```diff
@@ -121,7 +146,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 			_ = r.setProvisioningStatus(instance, newProvStatus)
 			return ctrl.Result{}, err
 		}
-		secretHash, op, err = common.CreateOrUpdateSecret(r, instance, deploymentSecret)
+		_, op, err = common.CreateOrUpdateSecret(r, instance, deploymentSecret)
```
stuggi, line 149: secretHash us used in L140, new L165, should we remove adding it to envVars ?

dprince: This is unrelated to this patch. I can remove this change perhaps. The reason this line is changed is I already tried removing adding it to envVars and it causes a variable reference error on compile. Perhaps we can follow up on a fix for this

dprince: Just pushed a version to drop this change...

pkg/common/passwords.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+type privateKey struct {
+	PrivateKey string `json:"private_key"`
+	PublicKey  string `json:"public_key"`
+}
+
+func generateSSHKeypair() (string, error) {
+
+	sshKey, err := rsa.GenerateKey(cryptorand.Reader, 2048)
```
dprince, line 212: Functionally this works fine. But I just realized we actually already have some code to generate ssh keys which we should try to reuse here. No harm in landing this and follow up to add that though.

controllers/openstackcontrolplane_controller.go
```diff
@@ -103,7 +103,32 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 	// Used in comparisons below to determine whether a status update is actually needed
 	newProvStatus := ospdirectorv1beta1.OpenStackControlPlaneProvisioningStatus{}
 
+	// ConfigMap - containing Tripleo Passwords
```
stuggi, line 106: it is Secret now

pkg/common/passwords.go
stuggi: agreed

stuggi (commented):
yes lgtm, only one comment references ConfigMap instead of Secret what it is now

controllers/openstackcontrolplane_controller.go
dprince: Fixed in latest commit

abays (approved):
/lgtm

PR #289, opened 2021-06-24
Make openstackclient CRD more generic

stuggi:
To use the openstackclient CRD for other purposes, like running
tempest using a tempest container image, the osclient CRD needs
to be more generic.

This makes DeploymentSSHSecret, GitSecret, and CloudName optional
parmeters and introduce RunUID and RunGID parameters which default
to the cloud-admin (42401) to be able to run pods using other/
custom users.

Also modifies the tripleo-deploy.sh to
* create ~/.config/openstack and copies the clouds.yaml there
instead of the current ~/tripleo-deploy directory. /home/cloud-admin
is a persistent storage so we can place it in the users home
directory. This resolves an issue where the openstackclient
pod gets recreated and the default /etc/openstack/clouds.yaml
is lost and would need to be copied from ~/tripleo-deploy/clouds.yaml

* adds ANSIBLE* environment variables for deployment like currently
set in an undercloud deployment script

### DIFF
--- a/api/v1beta1/openstackclient_types.go
+++ b/api/v1beta1/openstackclient_types.go
@@ -24,10 +24,16 @@ import (
 type OpenStackClientSpec struct {
 	// Name of the image
 	ImageURL string `json:"imageURL"`
+
+	// +kubebuilder:validation:Optional
 	// name of secret holding the stack-admin ssh keys
 	DeploymentSSHSecret string `json:"deploymentSSHSecret"`
+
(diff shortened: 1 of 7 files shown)


abays (commented):
+1 from my point of view.  Should probably let Dan have a look too.

stuggi:
just rebased

abays (approved):
/lgtm

PR #292, opened 2021-06-25
Use a secret for storing osephemeralheat password

dprince:


controllers/openstackephemeralheat_controller.go
```diff
@@ -96,13 +96,25 @@ func (r *OpenStackEphemeralHeatReconciler) Reconcile(ctx context.Context, req ct
...
+			r.Log.Info(fmt.Sprintf("Secret %s successfully reconciled - operation: ephemeral-heat-%s", instance.Name, string(op)))
+		}
+	}
+
 	envVars := make(map[string]common.EnvSetter)
 	templateParameters := make(map[string]interface{})
 	templateParameters["MariaDBHost"] = "mariadb-" + instance.Name
 	templateParameters["RabbitMQHost"] = "rabbitmq-" + instance.Name
-	templateParameters["MariaDBPassword"] = password
+	templateParameters["MariaDBPassword"] = passwordSecret.Data["password"]
```
dprince, line 117: This isn't quite right yet. Need to either convert to string and/or base64.decode?

abays: I think a base64 decode might do the trick

dprince: so I tried just adding string() and it seems to work fine. You do need to base64.decode if you access it from the CLI but apparently not via the Golang libraries

dprince:
Tested and this seems to functionally work fine.

stuggi (approved):
/lgtm

PR #293, opened 2021-06-25
KUTTL test for OSBaremetalSet scaling

abays:


config/samples/osp-director_v1beta1_openstackbaremetalset.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   # How many nodes to provision
   count: 1
   # The image to install on the provisioned nodes
-  baseImageUrl: http://172.22.0.1/images/rhel-guest-image-8.3-417.x86_64.qcow2.gz
+  baseImageUrl: http://192.168.111.1:8081/images/rhel-guest-image-8.4-825.x86_64.qcow2
```
dprince, line 10: my environment is actually on rhel-guest-image-8.4-1168.x86_64.qcow2 now. Is there a good way to make a parameter for this such that it works based on the image available locally in the environment

abays: Unfortunately, KUTTL doesn't provide that functionality at the moment.  Note that this is a change to our `OSBMS` sample YAML, and not specific to KUTTL itself.  So in a way, the samples are serving as configuration parameters.  We could perhaps write a shell script to run the tests that itself consumes environment variables or something, and then does some sort of string substitution before feeding the sample YAMLs to KUTTL.

abays:
Closing in favor of #295

PR #296, opened 2021-06-28
Status improvements for OSProvServer

abays:
This might be useful for KUTTL testing, or for manual status checks on the CLI


controllers/openstackprovisionserver_controller.go
```diff
@@ -194,9 +210,32 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
...
+		podList, err := r.Kclient.CoreV1().Pods(instance.Namespace).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("deployment=%s-provisionserver-deployment", instance.Name),
+		})
+
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			actualProvisioningState.State = ospdirectorv1beta1.ProvisionServerError
+			actualProvisioningState.Reason = err.Error()
+			_ = r.setProvisioningStatus(instance, actualProvisioningState)
+			return ctrl.Result{}, err
+		} else if !k8s_errors.IsNotFound(err) {
```
stuggi, line 224: do we need to check here on !k8s_errors.IsNotFound(err) ?

abays: I'm not sure what I was thinking.  I think it should just be `} else if err == nil {`.  If there's a "real" error, it is handled above the `else`, otherwise we want to proceed with the `else` as long as there isn't any error at all.

stuggi: do we want to catch/log hould we have k8s_errors.IsNotFound(err) situation in actualProvisioningState ?

~~~
if err != nil{
  if !k8s_errors.IsNotFound(err) {

  } else {

  }
  return ctrl.Result{}, err
} else {

}
~~~

stuggi (approved):
/lgtm

PR #301, opened 2021-06-30
Add finalizer to osephemeralheat

stuggi:
When the configmap in osplaybookgenerator changes, the
ephemeralheat gets recreated, but there is no finalizer used
yet to wait that all resource got deleted. When the resource
get not cleaned up correct, the password in the still running
mariadb pod and the heat pods can be different and the
playbookgeneration is broken.

This introduce a finalizer on osephemeralheat to cleanup all
resources properly when the osplaybookgenerator triggers a
recreate of the ephemeral heat stack.


pkg/common/pod.go
```diff
@@ -40,3 +43,29 @@ func GetAllPodsWithLabel(r ReconcilerCommon, labelSelectorMap map[string]string,
...
+	// delete all pods
+	for _, pod := range podList.Items {
+		err := r.GetClient().Delete(context.TODO(), &pod)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			err = fmt.Errorf("Error deleting pod %s: %v", pod.Name, err)
+			return err
+		}
+	}
+
+	return nil
```
abays, line 70: Just curious here: could a `r.Client.DeleteAllOf()` do the trick here without needing to get the list and then delete one-by-one?  I've never used the function myself.  Fine either way.

abays:
I added two thoughts, but really I think it's fine to merge as-is.

pkg/common/pod.go
stuggi: oh nice, wasn't aware of this. will check it!

stuggi: done

abays (approved):
/lgtm

PR #303, opened 2021-07-01
OSNet finalizer for NNCP removal

abays:
Adds a finalizer to handle the case when all OSNets specifying a particular bridge have been deleted.  If an OSNet is deleted and no other OSNets remain that target that bridge, the underlying NNCP will be removed.  NOTE: This does not handle the case where the user might change the name of the bridge in an OSNet.  We'll need to address that in another patch, one way or another.


stuggi (approved):
/lgtm

PR #305, opened 2021-07-02
Do not create CMs the osplaybookgenerator does not own

stuggi:


abays:
@stuggi Does this PR address the problem where the OSPG keeps restarting?

dprince (approved):

PR #308, opened 2021-07-07
Prevent OSNet bridge name change

abays:
Adds a webhook to disallow changing `attachConfiguration.desiredState`'s `interface`'s bridge name after the `OSNet` has been created.


api/v1beta1/openstacknet_types.go
```diff
@@ -161,3 +163,25 @@ type OpenStackNetList struct {
 func init() {
 	SchemeBuilder.Register(&OpenStackNet{}, &OpenStackNetList{})
 }
+
+// GetOpenStackNetAttachConfigBridgeName - Get the bridge name associated with the desiredState of the attachConfiguration
+func (openstackNet *OpenStackNet) GetOpenStackNetAttachConfigBridgeName() (string, error) {
```
stuggi, line 168: can we put this in pkg/openstacknet/funcs.go  where we also have GetOpenStackNetsAttachConfigBridgeNames?

abays: That might be a possibility.  I was waiting until we had #303 merged so that I could try to weave it into that.  The only concern is potential circular dependencies, but let's see what happens.

abays: Yeah, it does cause a circular import.  Let me see what else I can do to make it `DRY`.

stuggi (approved):
/lgtm

PR #318, opened 2021-07-13
Use consolidated img URLs in KUTTL

abays:
Adjusts KUTTL tests to use the shared URL structure for RHEL image locations

Depends-on: https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/172


stuggi (approved):
/lgtm

PR #319, opened 2021-07-13
Change order in stack created command to overwrite password

stuggi:
This changes the order of the password environment file provided
to the stack created command. With this it is possible to customize
passwords for testing via the TripleoCustomDeployFiles.


abays (approved):
/lgtm

PR #321, opened 2021-07-13
Set deploymentSSHSecret: osp-controlplane-ssh-keys

dprince:
This key is automatically generated by the controlplane AppLabel
name. Lets make them match.


stuggi (commented):
this change needs also be done in the kuttl test files

dprince:
> this change needs also be done in the kuttl test files

interesting. I would not have expected it to work correctly as the kuttle tests are currently coded as that secret won't even exist. I guess it just checks the API though so it makes sense

stuggi (approved):
/lgtm

PR #322, opened 2021-07-14
Generate tripleo-passwords only once

stuggi:
Right now on every reconcile the tripleo-passwords get created
and the secret updated. The tripleo-passwords should only be created
once if the secret don't exist.


abays (approved):
/lgtm

PR #333, opened 2021-07-19
Add label for must-gather-safe secrets

abays:
Adds a label to network data secrets in `VMSets` and `BMSets` to indicate that they are safe to collect via `must-gather`.


stuggi (approved):
/lgtm

PR #344, opened 2021-07-27
Update create-playbooks.sh for new net-config

dprince:
This updates the create-playbooks.sh script so that it skips customizing
run-os-net-config.sh script if it doesn't exist (it no longer exists in OSP17+).

Instead users can now set the NetworkSafeDefaults Heat ENV variable to
influence the os_net_config tripleo ansible module to control the
same setting.


abays (commented):
Seems fine from a bash perspective, but I can't really speak to the OOO changes

PR #346, opened 2021-07-29
Better OsProvServer port handling

abays:
To prevent duplicate `OpenStackProvisionServer` Apache listening ports:

1. Adds webhook for `OpenStackProvisionServer` to prevent explicit duplicates from user CR create/apply
2. Adds port auto-assign logic for `OpenStackProvisionServer` for CRs generated by `OpenStackBaremetalSet` controller

Also modifies the Apache containers to disable SSL, as it is not needed and just invites other port conflicts


api/v1beta1/openstackprovisionserver_types.go
```diff
@@ -17,7 +17,10 @@ limitations under the License.
 package v1beta1
 
 import (
+	"context"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goClient "sigs.k8s.io/controller-runtime/pkg/client"
```
abays, line 23: The client is imported and used here because the added functions are needed by both the webhook and the controllers.  To avoid import cycles and also allow the operator to run locally (without webhooks, and therefore without `webhookClient`, passing a client seemed like a reasonable option.  We can change this is someone has a better idea.

dprince (approved):

PR #361, opened 2021-08-17
go mod tidy

dprince:
Drop a conflict with grpc/examples (caused an error).

Also drops the now unused ghodss/yaml


abays (approved):
/lgtm

PR #366, opened 2021-08-30
Drop SRIOV version dep from provserver container

dprince:
This drops the sriov version dependency and log
statement from the provisionserver container
(shouldn't be needed)

### DIFF
--- a/containers/provision_ip_discovery_agent/start.go
+++ b/containers/provision_ip_discovery_agent/start.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/openshift/sriov-network-operator/pkg/version"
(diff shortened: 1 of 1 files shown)


containers/provision_ip_discovery_agent/start.go
```diff
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/openshift/sriov-network-operator/pkg/version"
```
dprince, line 12: @abays Is the intent to use the version.go from this directory instead?

dprince: that appears to be something different though, perhaps we just remove this

abays (approved):
/lgtm

PR #367, opened 2021-08-30
Use RerunOnFailure Kubevirt run strategy for VMs

abays:
As explain in code comments, the Kubevirt fencing agent needs to be able to directly call the Kubevirt API for power operators.  This is not allowed by the default `Always` run strategy.  Instead, we need to use `RerunOnFailure` which will still boot/reboot the VM on creation/failure, but also allows for direct power API calls.


stuggi (approved):
/lgtm

PR #372, opened 2021-09-02
provisionserver: default to rhel8/httpd-24:1-152

dprince:
Update the OpenStackProvisionServer defaults so that we use
registry.redhat.io/rhel8/httpd-24:1-152 as our default image.

Also fixes a functional issue with this image so that ServerRoot
is set correctly.


abays (approved):
/lgtm

PR #373, opened 2021-09-03
Use 16.2 instead of more specific tags

dprince:
This should act as the 'latest' 16.2 and is what we default
to in containers-image-prepare via tripleoclient so it should be
a better default.

Also, use the 'latest' tag for the rhel8/httpd-24.


stuggi (approved):
/lgtm

PR #374, opened 2021-09-03
Use 'latest' tag for http image default

dprince:


config/crd/bases/osp-director.openstack.org_openstackprovisionservers.yaml
```diff
@@ -51,7 +51,7 @@ spec:
               OpenStackProvisionServer
             properties:
               apacheImageUrl:
-                default: registry.redhat.io/rhel8/httpd-24:1-152
+                default: registry.redhat.io/rhel8/httpd-24:latest
```
stuggi, line 54: when using latest, should we set `imagePullPolicy: Always` in the controller creating the osp-httpd container?

dprince: yes! Just added a new commit to address this

stuggi (approved):
/lgtm

PR #382, opened 2021-09-13
Generate fencing YAML

abays:


controllers/openstackplaybookgenerator_controller.go
```diff
@@ -114,6 +110,89 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
...
+
+				if err != nil {
+					_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+					return ctrl.Result{}, err
+				}
+
+				templateParameters = common.MergeTemplateParameters(templateParameters, fencingTemplateParameters)
+				templateParameters["EnableFencing"] = true
+				// Only should be one Controller role, so break
+				break
```
stuggi, line 168: if you e.g. use ControllerOpenstack, which does not run rabbit and db, you'd also have VMs for Database and Messaging roles which also required fencing to be enabled. I know this is to add initial fencing config, just to keep in mind that if we have a role with 3 nodes+pacemaker service there might be also others.

pkg/controlplane/funcs.go
```diff
@@ -0,0 +1,21 @@
...
+		"ControllerNovaStandalone",
+		"ControllerOpenstack",
+		"ControllerSriov",
+		"ControllerStorageDashboard",
+		"ControllerStorageNfs",
+		"Controller",
+		"Database",
+		"Messaging",
+		"Standalone",
+		"Telemetry",
```
stuggi, line 19: we could also get information from the default roles from inside the openstackclient image from roles in /usr/share/openstack-tripleo-heat-templates/roles

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -114,6 +110,89 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
...
+		err = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorWaiting, msg)
+		return ctrl.Result{RequeueAfter: time.Second * 10}, err
+	}
+
+	templateParameters["EnableFencing"] = false
+
+	if controlPlane.Spec.EnableFencing {
+		// TODO: This will likely need refactoring sooner or later
+		for roleName, roleParams := range controlPlane.Spec.VirtualMachineRoles {
+			if common.StringInSlice(roleParams.RoleName, controlplane.GetFencingRoles()) && roleParams.RoleCount == 3 {
```
stuggi, line 147: IIUC you want to address custom roles where the user provided a roles file via the tripleo-tarball-config configmap in a follow up, right?

pkg/controlplane/configmap.go
```diff
@@ -0,0 +1,59 @@
...
+
+		fencingConfigs = append(fencingConfigs, fencingConfig)
+	}
+
+	// Sort the config so that any generated hashes remain constant (we don't know the order
+	// of the virtualMachineInstanceList)
+
+	sort.Slice(fencingConfigs, func(i, j int) bool {
+		return fencingConfigs[i].VMName < fencingConfigs[j].VMName
+	})
```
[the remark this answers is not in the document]
abays: Sure.  If there's a good way to abstract it, let's do it.

controllers/openstackplaybookgenerator_controller.go
abays: Ack, thanks for the feedback on this point.  Perhaps we remove the break and let the loop generate fencing device data for all VMs that belong to any role in `common/GetFencingRoles()`?

stuggi: yes, for the common.GetFencingRoles() it would be fine + checking for roles having `OS::TripleO::Services::Pacemaker` in any provided custom roles_data.yaml

controllers/openstackplaybookgenerator_controller.go
abays: Yes, correct.  Will have to look into unpacking binary CM data inside the operator and parsing through it to find the potentially relevant role chunks.

pkg/controlplane/funcs.go
abays: That would be cool.  What would be the approach?  I'm just wondering what you had in mind for how the operator would access the data inside the openstackclient image or pod.

pkg/controlplane/configmap.go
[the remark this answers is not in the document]
abays: I was thinking about this again.  Do you want to just put a wrapper around it and place it in the common functions?  The reason I ask is because it seems to be as abstracted as it can be as-is.  i.e. we'd have to pass the struct slice and a sorting function to the wrapper anyhow.  Or is there something else you had in mind?

pkg/controlplane/funcs.go
stuggi: When the ctlplane CR gets created we can run an init job, which gets OSP information from the openstackclient image into into a CM. the controllers which need default ooo information could then parse the information from the cm.

pkg/common/template_util.go
```diff
@@ -129,6 +129,17 @@ func ExecuteTemplateFile(filename string, data interface{}) string {
 	return buff.String()
 }
 
+// MergeTemplateParameters - takes an list of template parameter maps and merges them into an existing template parameter map
+func MergeTemplateParameters(templateParameters map[string]interface{}, extraTemplateParameters ...map[string]interface{}) map[string]interface{} {
```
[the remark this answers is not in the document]
abays: Sure, sounds good

controllers/openstackplaybookgenerator_controller.go
abays: Can we do the checking of the provided custom roles in another follow-up PR?  I'd just like to get this one merged and then we can build on top of it.

pkg/controlplane/funcs.go
abays: Alright, let's also do this in a follow-up PR if you would be okay with that

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -114,6 +111,93 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
...
+	err = r.Client.List(context.TODO(), controlPlaneList, controlPlaneListOpts...)
+	if err != nil {
+		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	if len(controlPlaneList.Items) == 0 {
+		err := fmt.Errorf("no OpenStackControlPlanes found in namespace %s", instance.Namespace)
+		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
abays: Yes, good catch

pkg/controlplane/configmap.go
```diff
@@ -0,0 +1,61 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controlplane
+
+import (
+	//	common "github.com/openstack-k8s-operators/osp-director-operator/pkg/common"
```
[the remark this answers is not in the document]
abays: Ack

controllers/openstackplaybookgenerator_controller.go
stuggi: sure

pkg/controlplane/funcs.go
stuggi: ack

pkg/common/template_util.go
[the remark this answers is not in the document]
stuggi: this can be removed now as its no longer used, right?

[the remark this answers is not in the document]
abays: Ack, will fix

stuggi (approved):
/lgtm

PR #383, opened 2021-09-14
Update kuttl tests to latest env changes

stuggi:
Modify kuttl tests to:
- use updated images
- br-ex for external network
- update interfaces to match dev-tools interface mapping

Also updates kuttl config to make sure no tests run in parallel
(default 8) as this result in failures.


abays (approved):
/lgtm

PR #384, opened 2021-09-16
Update OSP images to release version with 16.2 tag

dprince:
NOTE: I have pinned the tripleoclient to 16.2.0 instead
of 16.2 because that tag doesn't yet exist. See also:
https://bugzilla.redhat.com/show_bug.cgi?id=2005107


abays (approved):
/lgtm

PR #386, opened 2021-09-18
Fencing SA and associated kubeconfig

abays:
Adds a service account for the Kubevirt fencing functionality, and then generates and injects a kubeconfig for that account into VMSet VMs to be used with the fencing agent

### DIFF
--- a/controllers/openstackvmset_controller.go
+++ b/controllers/openstackvmset_controller.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/utils/diff"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	//networkv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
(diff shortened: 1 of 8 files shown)


controllers/openstackvmset_controller.go
```diff
@@ -675,6 +690,66 @@ func (r *OpenStackVMSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
+
+	tokenSecretName := ""
+
+	for _, secret := range kubevirtAgentServiceAccount.Secrets {
+		if strings.Contains(secret.Name, fmt.Sprintf("%s-token", vmset.KubevirtFencingServiceAccount)) {
+			tokenSecretName = secret.Name
+			break
+		}
+	}
```
stuggi, line 719: like we discussed a last week in slack, we could also create our dedicated named token for the SA [1]
so the we could just get this specific token instead of searching for any of the auto created tokens for
the SA. Do the auto generated ones change?

[1] https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#to-create-additional-api-tokens

abays: I looked into this initially and discovered that it just creates a duplicate of the `osp-director-operator-kubevirtagent-token-XXXXX` secret, so that why I just find those secrets and get the token from there.  They don't appear to change unless you uninstall/reinstall the CSV.  But you are right, it is simpler perhaps to just create the duplicate with a known name, instead of getting the SA and then finding the right secret.  I'll rework this piece.

dprince (approved):

PR #389, opened 2021-09-21
Updated jenkins pipeline to include branch name

mpryc:
Updated pipeline file that include GITHUB_BRANCH variable.
Unfortunately Jenkins multibranch do not pass the target
branch name, so this file will need to be kept updated
for every branch containing fixed name.


stuggi (approved):
/lgtm

PR #390, opened 2021-09-21
Kubeconfig for fencing, part 2

abays:
1. OSVMSet controller creates namespace-scoped secrets needed for fencing considerations, instead of relying on OLM (which had a race condition)
2. Allows config maps and secrets to have annotations
3. Allows secrets to set a secret type (instead of only Opaque)
4. Adds functionality to unzip custom heat tarballs to detect custom roles that use Pacemaker

### DIFF
--- a/controllers/openstackplaybookgenerator_controller.go
+++ b/controllers/openstackplaybookgenerator_controller.go
@@ -111,6 +111,85 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 	templateParameters := make(map[string]interface{})
 	cmLabels := common.GetLabels(instance, openstackplaybookgenerator.AppLabel, map[string]string{})
 
+	// check if heat-env-config (customizations provided by administrator) exist
+	// if it does not exist, requeue
+	tripleoCustomDeployCM, _, err := common.GetConfigMapAndHashWithName(r, instance.Spec.HeatEnvConfigMap, instance.Namespace)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			msg := fmt.Sprintf("The ConfigMap %s doesn't yet exist. Requeing...", instance.Spec.HeatEnvConfigMap)
+			r.Log.Info(msg)
+			err = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorWaiting, msg)
+			return ctrl.Result{RequeueAfter: time.Second * 10}, err
+		}
+		// Ignore any potential error from "setCurrentState"; focus on previous error
+		// NOTE: This pattern will be repeated a lot below, but the comment will not be copied
+		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	tripleoCustomDeployFiles := tripleoCustomDeployCM.Data
+	templateParameters["TripleoCustomDeployFiles"] = tripleoCustomDeployFiles
+
+	// Now read the tripleo-deploy-config and the fencing-config CMs for use in the PlaybookGenerator
+	tripleoDeployCM, _, err := common.GetConfigMapAndHashWithName(r, "tripleo-deploy-config", instance.Namespace)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			msg := "The tripleo-deploy-config map doesn't yet exist. Requeing..."
+			r.Log.Info(msg)
+			err = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorWaiting, msg)
+			return ctrl.Result{RequeueAfter: time.Second * 10}, err
+		}
+		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	tripleoDeployFiles := tripleoDeployCM.Data
+	// Also add fencing.yaml to the tripleoDeployFiles (just need the file name)
+	templateParameters["TripleoDeployFiles"] = tripleoDeployFiles
+	templateParameters["HeatServiceName"] = "heat-" + instance.Name
+
+	var tripleoTarballCM *corev1.ConfigMap
+
+	// extra stuff if custom tarballs are provided
+	if instance.Spec.TarballConfigMap != "" {
+		tripleoTarballCM, _, err = common.GetConfigMapAndHashWithName(r, instance.Spec.TarballConfigMap, instance.Namespace)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				err = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorWaiting, "Tarball config map not found, requeuing and waiting")
+				return ctrl.Result{RequeueAfter: time.Second * 10}, err
+			}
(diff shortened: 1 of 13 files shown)


controllers/openstackvmset_controller.go
```diff
@@ -852,6 +873,8 @@ func (r *OpenStackVMSetReconciler) vmCreateInstance(instance *ospdirectorv1beta1
 	fsMode := corev1.PersistentVolumeFilesystem
 	trueValue := true
 	terminationGracePeriodSeconds := int64(0)
+	// TODO: How do we account for custom roles here?
```
stuggi, line 876: do you mean custom tripleo roles? or VMs we just prepare but are not part of tripleo deployment?

templates/openstackplaybookgenerator/config/fencing.yaml
```diff
@@ -13,6 +13,6 @@ parameter_defaults:
           namespace: {{ $fencingConfig.Namespace }}
           ssl_insecure: true
           power_timeout: 30
-          kubeconfig: /home/cloud-admin/kubeconfig
+          kubeconfig: /mnt/kubeconfig/kubeconfig
```
stuggi, line 16: not related to this file, but to the kubeconfig in general:
* what are the permissions to the kubeconfig? we need to make sure that only root can access it so that it can not be used by any other user. maybe we just place it in /root/kubeconfig/kubeconfig ?
* how does pacemaker behave if the config is no longer accessible for whatever reason. should we copy the config on boot instead of direct use of the secret mount?

controllers/openstackvmset_controller.go
abays: Yes, custom OOO roles.  In the playbook generator, we have the custom tarball and can now crack it open and examine `roles_data.yaml`, if any.  In that case of VMSets though, we don't have a window into these potential custom roles.  Maybe we should just always mount the minimal kubeconfig for Pacemaker and make sure its filesystem perms are as restrictive as possible (as you suggest in your other comment)?

templates/openstackplaybookgenerator/config/fencing.yaml
abays: We can certainly mount the kubeconfig elsewhere.  And yes, we can also tighten the filesystem perms to `0600` for `root`-only, or something like that.

If the kubeconfig is not available, the call to `/usr/sbin/fence_kubevirt` will fail.  In terms of what you'd see in the OSP cloud, `pcs status` would report a failed fencing action.

templates/vmset/cloudinit/userdata
```diff
@@ -12,4 +12,8 @@ chpasswd:
 {{- end }}
 bootcmd:
   - "mkdir /mnt/kubeconfig"
-  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
\ No newline at end of file
+  - "mkdir /root/kubeconfig"
```
stuggi, line 15: bootcmd will run on every boot. should be `mkdir -p` for later reboots, same for the above mkdir, or just use single `mkdir -p /mnt/kubeconfig /root/kubeconfig`

templates/vmset/cloudinit/userdata
```diff
@@ -12,4 +12,8 @@ chpasswd:
...
+  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
+  - "chown -R root:root /mnt/kubeconfig"
+  - "chmod -R 0600 /mnt/kubeconfig"
+  - "cp /mnt/kubeconfig/* /root/kubeconfig/."
```
stuggi, line 19: bootcmd will run on every boot, so we should do `cp -f` to update existing kubeconfig on later reboots

templates/vmset/cloudinit/userdata
```diff
@@ -12,4 +12,8 @@ chpasswd:
 {{- end }}
 bootcmd:
   - "mkdir /mnt/kubeconfig"
-  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
\ No newline at end of file
+  - "mkdir /root/kubeconfig"
+  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
+  - "chown -R root:root /mnt/kubeconfig"
+  - "chmod -R 0600 /mnt/kubeconfig"
```
stuggi, line 18: could also just umount after the successful cp

templates/vmset/cloudinit/userdata
```diff
@@ -12,4 +12,8 @@ chpasswd:
 {{- end }}
 bootcmd:
   - "mkdir /mnt/kubeconfig"
-  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
\ No newline at end of file
+  - "mkdir /root/kubeconfig"
+  - "mount /dev/disk/by-id/virtio-fencingdisk /mnt/kubeconfig"
```
stuggi, line 16: could also check to use the mounts feature of cloud-init https://cloudinit.readthedocs.io/en/latest/topics/examples.html#adjust-mount-points-mounted

controllers/openstackvmset_controller.go
stuggi: yes, as discussed on slack

templates/openstackplaybookgenerator/config/fencing.yaml
stuggi: as discussed on slack, that was the idea on this to prevent any fencing action because the kubeconfig is not available.

templates/vmset/cloudinit/userdata
abays: I'll take a look

templates/vmset/cloudinit/userdata
abays: Ack

templates/vmset/cloudinit/userdata
abays: Decided to just keep it as-is and add your flag and unmount suggestions

stuggi (commented):
in general looks good to me now. it just needs a rebase on latest master

abays:
The rebase went haywire.  Trying to fix.

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -111,6 +111,85 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
...
+		// NOTE: This pattern will be repeated a lot below, but the comment will not be copied
+		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	tripleoCustomDeployFiles := tripleoCustomDeployCM.Data
+	templateParameters["TripleoCustomDeployFiles"] = tripleoCustomDeployFiles
+
+	// Now read the tripleo-deploy-config and the fencing-config CMs for use in the PlaybookGenerator
+	tripleoDeployCM, _, err := common.GetConfigMapAndHashWithName(r, "tripleo-deploy-config", instance.Namespace)
```
stuggi, line 134: just realized we do not have a const for tripleo-deploy-config and use it now in couple of places. Might be good we change that in a later change.

stuggi (approved):
/lgtm

controllers/openstackplaybookgenerator_controller.go
stuggi: depending on how we implement DCN that can't be a const... so lets wait for the work on that feature

PR #391, opened 2021-09-21
Set default images to use OSP 16.2 release versions

dprince:


stuggi (approved):
/lgtm we also have to update the assert files in the kuttl tests for the images references. but we can do that in a follow up

PR #393, opened 2021-09-22
OSPK8-352 Support rendering required parameter files for osp16 and 17

stuggi:
The osctlplane controller checks the tripleoclient image url and
gets the OSP version from the image tag. Expected to have either
16.2 or 17.0. Everything after that tag prefix get ignored.

The OSP version gets recorded in the CR status of the osctlplane.
Other controllers which require this information can get the OSP
version from there. Right now this is the osipset and
osplaybookgeneratior.

Also updates create-playbooks.sh to match the inventory file name
and rendered playbook location to match what is done in 17.


controllers/openstackplaybookgenerator_controller.go
```diff
@@ -111,27 +111,12 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
...
 	if err != nil {
-		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
-		return ctrl.Result{}, err
-	}
-
-	if len(controlPlaneList.Items) == 0 {
-		err := fmt.Errorf("no OpenStackControlPlanes found in namespace %s. Requeing", instance.Namespace)
-		_ = r.setCurrentState(instance, ospdirectorv1beta1.PlaybookGeneratorError, err.Error())
-		return ctrl.Result{RequeueAfter: time.Second * 10}, err
+		return ctrlResult, err
```
abays, line 117: Should we set the status here in the case of an error?

controllers/openstackcontrolplane_controller.go
```diff
@@ -44,6 +44,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// OSPVersion - global var to specify tge OSP version, defaults to OSP17
+var OSPVersion = ospdirectorv1beta1.TemplateVersionWallaby
```
dprince, line 48: I think we should consider defaulting to OSP 16.2 for now to match other settings in the Operator.

abays (commented):
One question from me and another from Dan, but otherwise I think this is ready

controllers/openstackplaybookgenerator_controller.go
stuggi: done

controllers/openstackcontrolplane_controller.go
stuggi: done

abays (approved):
/lgtm

PR #395, opened 2021-09-24
OSPK8-358 update kuttl tests

stuggi:
* assert for default switch to 16.2
* unique_role_name test to deploy an controlplane as this is now mandatory for a vmset


abays (approved):
/lgtm

PR #396, opened 2021-09-24
OSPK8-342 Add support for setting fqdn

olliewalsh:
Adds a domainName field to the CRDs. If set then the openstackclient pod
and VM/BM hostnames will be fully qualified.


stuggi:
should this also add CloudDomain/CloudName* parameters to the env file the operator renders and what our dev-tools does in [1]? or should the CloudDomain parameter added to a cm like [1] be the source for input and sets the CRD parameter?

```
parameter_defaults:
  CloudDomain: {{ ocp_cluster_name }}
  CloudName: overcloud.{{ ocp_cluster_name }}
  CloudNameInternal: overcloud.internalapi.{{ ocp_cluster_name }}
  CloudNameStorage: overcloud.storage.{{ ocp_cluster_name }}
  CloudNameStorageManagement: overcloud.storagemgmt.{{ ocp_cluster_name }}
  CloudNameCtlplane: overcloud.ctlplane.{{ ocp_cluster_name }}

```

[1] https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/master/ansible/templates/osp/tripleo_heat_envs/common/cloud-names.yaml.j2

olliewalsh:
> should this also add CloudDomain/CloudName* parameters to the env file the operator renders and what our dev-tools does in [1]?
Yes, could generate CloudDomain* based on this. Probably should make DomainName a required param in that case.

> or should the CloudDomain parameter added to a cm like [1] be the source for input and sets the CRD parameter?
Is that an option? Does the cm exists yet?

controllers/openstackcontrolplane_controller.go
```diff
@@ -364,6 +367,9 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 		osc.Spec.GitSecret = instance.Spec.GitSecret
 		osc.Spec.RunUID = openstackclient.CloudAdminUID
 		osc.Spec.RunGID = openstackclient.CloudAdminGID
+		if instance.Spec.DomainName != "" {
+			osc.Spec.DomainName = instance.Spec.DomainName
+		}
```
olliewalsh, line 372: TODO: how to pass the domainname through to the BMset

abays: Might just need to use the new `common.GetControlPlane` func within the BaremetalSet controller to get access to the associated ControlPlane resource

abays (commented):
Seems ready for merge from my perspective.  Will let @stuggi apply the flag if he agrees

PR #398, opened 2021-09-29
Fix condition for rendered parameter file

stuggi:
Since 16.2 and 17 is supported on master branch, same file gets
rendered and condition is wrong.


abays (approved):
/lgtm

PR #404, opened 2021-10-05
Fix default tht dir replacement

stuggi:
Make sure /usr/share/openstack-tripleo-heat-templates is also removed
in any file added via the tht tar file. E.g. in the provided nic templates.

### DIFF
--- a/templates/openstackplaybookgenerator/bin/create-playbooks.sh
+++ b/templates/openstackplaybookgenerator/bin/create-playbooks.sh
@@ -38,25 +38,26 @@ rm -Rf "$TEMPLATES_DIR"
 mkdir -p $TEMPLATES_DIR
 
 cp -a /usr/share/openstack-tripleo-heat-templates/* $TEMPLATES_DIR
-pushd $TEMPLATES_DIR
-# extract any tar files into the $TEMPLATES_DIR
+
+# copy to editable dir config-tmp
+rm -Rf $HOME/config-tmp
+mkdir -p $HOME/config-tmp
+pushd $HOME/config-tmp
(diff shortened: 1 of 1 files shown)


templates/openstackplaybookgenerator/bin/create-playbooks.sh
```diff
@@ -38,17 +38,17 @@ rm -Rf "$TEMPLATES_DIR"
 mkdir -p $TEMPLATES_DIR
 
 cp -a /usr/share/openstack-tripleo-heat-templates/* $TEMPLATES_DIR
-pushd $TEMPLATES_DIR
-# extract any tar files into the $TEMPLATES_DIR
+
+# copy to editable dir config-tmp
+rm -Rf $HOME/config-tmp
```
dprince, line 43: Why do you need to remove it? Shouldn't it always not exist?

stuggi: as discussed on slack we should keep that when running in interactive mode you want to rerun the script

dprince:
@stuggi So the only reason this works is because we are doing the following to config-tmp right?

#FIXME: get rid of /usr/share/openstack-tripleo-heat-templates/ and use relative paths in dev-tools!
sed -e "s|/usr/share/openstack\-tripleo\-heat\-templates|\.|" -i $HOME/config-tmp/*.yaml

I think I wrote ^^ comment and my take was always that we should stop using hard coded /usr/share/openstack paths in our templates. Based on the fact that you are now relying on that as a feature should we remove or update the comment to reflect that you are now depending on this sed statement as a feature.

dprince (approved):

PR #411, opened 2021-10-11
OSPK8-371 IPv6 support

stuggi:
- Rename overcloudipset package to openstackipset
- Correct renders IPAddrURI if a network is IPv6 network.
- Disable ipv6 autoconfig networkmanager firstboot
    On initial boot set NM ipv6 connection method to link-local
    to remove all IPs which might have been configured via SLAAC.
    This is especially required for interface where VIPs get created.
    The puppt module might otherwise pick a wrong interface and fail,
    e.g external interface, where the VIP should create on br-ex.
    With the oo deployment we switch to network-scripts.
    
    Also disables IPv6 autoconf per default, we only want the ooo
    static IP config.
    
    Both is limited to VMs which are part of a ooo role.

- Add support for upstream tripleoclient container image
    Until all patches backported to upstream wallaby there are two possibilities:
       -quay.io/tripleowallaby/openstack-tripleoclient:current-tripleo
       -quay.io/openstack-k8s-operators/tripleowallaby-openstack-tripleoclient:current-tripleo_patched

- Also adds support to deploy upstream train/wallaby release

### DIFF
--- a/pkg/common/container.go
+++ b/pkg/common/container.go
@@ -48,8 +48,8 @@ func GetOSPVersion(parsedVersion string) (ospdirectorv1beta1.OSPVersion, error) 
 	}
 }
 
-// GetVersionFromImageURL -
-func GetVersionFromImageURL(r ReconcilerCommon, imageURL string) (ospdirectorv1beta1.OSPVersion, error) {
+// GetOSPVersionFromImageURL -
+func GetOSPVersionFromImageURL(r ReconcilerCommon, imageURL string) (ospdirectorv1beta1.OSPVersion, error) {
 
 	parts := strings.SplitN(imageURL, ":", 2)
 	if len(parts) != 2 {
--- a/controllers/openstackcontrolplane_controller.go
+++ b/controllers/openstackcontrolplane_controller.go
@@ -115,7 +115,7 @@ func (r *OpenStackControlPlaneReconciler) Reconcile(ctx context.Context, req ctr
 	if instance.Spec.OpenStackRelease != "" {
 		OSPVersion, err = common.GetOSPVersion(instance.Spec.OpenStackRelease)
 	} else {
-		OSPVersion, err = common.GetVersionFromImageURL(r, instance.Spec.OpenStackClientImageURL)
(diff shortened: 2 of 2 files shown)


pkg/common/container.go
```diff
@@ -41,21 +49,36 @@ func getOSPVersion(parsedVersion string) (ospdirectorv1beta1.OSPVersion, error)
 }
 
 // GetVersionFromImageURL -
-func GetVersionFromImageURL(imageURL string) (ospdirectorv1beta1.OSPVersion, error) {
+func GetVersionFromImageURL(r ReconcilerCommon, imageURL string) (ospdirectorv1beta1.OSPVersion, error) {
```
abays, line 52: Not important, but a suggestion: maybe call it `GetOSPVersionFromImageURL` just to be more specific on what the version refers to?

stuggi: makes sense, done

pkg/common/container.go
```diff
@@ -23,39 +23,62 @@ import (
...
 	match := re.FindStringSubmatch(tag)
 
-	// verify parsed OSP version
-	ospVersionm, err := getOSPVersion(match[1])
+	// if no match in checking image tag, check for upstream image version in URL
+	if match == nil || match[1] == "" {
+		// Until all patches backported to upstream wallaby there are two possibilities:
+		// -quay.io/tripleowallaby/openstack-tripleoclient:current-tripleo
+		// -quay.io/openstack-k8s-operators/tripleowallaby-openstack-tripleoclient:current-tripleo_patched
+		re = regexp.MustCompile(`.*\/tripleo(.*)[-\/](openstack-tripleoclient).*`)
```
dprince, line 72: the logic in this regex strikes me as weird. Is there another way to handle this perhaps with just a flag?

stuggi: Sorry I think I do not get that comment on using a flag. This PR also adds a OpenStackRelease parameter to the ctlplane where you can always overwrite auto discovery, if that is what you meant.

dprince: What I mean is maybe we shouldn't be using the Image URL as the mechanism to determine which version of OSP is getting used.

dprince:
This looks good to me.

PR #412, opened 2021-10-13
Generate OpenStackConfigVersions

dprince:
OpenStackConfig versions is a new API to track git commits and diffs vs 'latest' with an operator native workflow. This patch generates openstackconfigversion objects once playbook generation finishes.

The plan is to use this to drive a future OpenStackDeployment CRD and thus remove the need for users to run tripleo-deploy.sh in a pod.

### DIFF
--- a/controllers/openstackplaybookgenerator_controller.go
+++ b/controllers/openstackplaybookgenerator_controller.go
@@ -466,7 +466,7 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 	configVersions, gerr := openstackconfigversion.SyncGit(instance, r.Client, r.Log)
 	if gerr != nil {
 		r.Log.Error(gerr, "ConfigVersions")
-		return ctrl.Result{}, err
+		return ctrl.Result{}, gerr
 	}
 
 	if err := r.syncConfigVersions(instance, configVersions); err != nil {


controllers/openstackconfigversion_controller.go
```diff
@@ -0,0 +1,63 @@
...
+//+kubebuilder:rbac:groups=osp-director.openstack.org,resources=openstackconfigversions/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the OpenStackConfigVersion object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
```
dprince, line 48: Not really using this yet. One idea would be to use this to implement a finalizer on OpenStackConfig objects where we attempt to delete git commits if the osconfigversion gets deleted. But that might be overkill?

pkg/openstackconfigversion/git_util.go
```diff
@@ -0,0 +1,221 @@
...
+// SyncGit func
+func SyncGit(inst *ospdirectorv1beta1.OpenStackPlaybookGenerator, client client.Client, log logr.Logger) (map[string]ospdirectorv1beta1.OpenStackConfigVersion, error) {
+
+	configVersions := make(map[string]ospdirectorv1beta1.OpenStackConfigVersion)
+
+	// Check if this Job already exists
+	foundSecret := &corev1.Secret{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: inst.Spec.GitSecret, Namespace: inst.Namespace}, foundSecret)
+	if err != nil {
+		log.Info("GitRepo err")
```
stuggi, line 120: this message should probably  be more specific to say that the inst.Spec.GitSecret secret was not found instead of  this generic error message?

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -495,6 +508,29 @@ func (r *OpenStackPlaybookGeneratorReconciler) setCurrentState(instance *ospdire
 
 }
 
+func (r *OpenStackPlaybookGeneratorReconciler) syncConfigVersions(instance *ospdirectorv1beta1.OpenStackPlaybookGenerator, configVersions map[string]ospdirectorv1beta1.OpenStackConfigVersion) error {
+
+	for _, version := range configVersions {
+
+		// Check if this Route already exists
```
stuggi, line 515: Route => ConfigVersion?

pkg/openstackconfigversion/git_util.go
```diff
@@ -0,0 +1,221 @@
...
+
+	return retPatches
+}
+
+// SyncGit func
+func SyncGit(inst *ospdirectorv1beta1.OpenStackPlaybookGenerator, client client.Client, log logr.Logger) (map[string]ospdirectorv1beta1.OpenStackConfigVersion, error) {
+
+	configVersions := make(map[string]ospdirectorv1beta1.OpenStackConfigVersion)
+
+	// Check if this Job already exists
```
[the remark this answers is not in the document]
dprince: got it

config/crd/bases/osp-director.openstack.org_openstackconfigversions.yaml
```diff
@@ -0,0 +1,66 @@
...
+  creationTimestamp: null
+  name: openstackconfigversions.osp-director.openstack.org
+spec:
+  group: osp-director.openstack.org
+  names:
+    kind: OpenStackConfigVersion
+    listKind: OpenStackConfigVersionList
+    plural: openstackconfigversions
+    shortNames:
+    - osipset
```
olliewalsh, line 17: typo?

bundle.Dockerfile
```diff
@@ -6,8 +6,8 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=osp-director-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.11.0+git
```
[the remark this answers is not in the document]
dprince: so I was trying it out and it seems to work fine. I do think it is time for us to bump up (we are still on 1.5)

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -495,6 +508,29 @@ func (r *OpenStackPlaybookGeneratorReconciler) setCurrentState(instance *ospdire
...
+			r.Log.Info("Creating a ConfigVersion", "ConfigVersion.Namespace", instance.Namespace, "ConfigVersion.Name", version.Name)
+			err = r.Client.Create(context.TODO(), &version)
+			if err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
+	}
+	//FIXME(dprince): remove deleted?
```
abays, line 530: Perhaps the same consideration here as with regards to the finalizer conversation above

controllers/openstackplaybookgenerator_controller.go
```diff
@@ -460,6 +462,17 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 		return ctrl.Result{}, err
 	}
 
+	// update ConfigVersions with Git Commits
+	configVersions, gerr := openstackconfigversion.SyncGit(instance, r.Client, r.Log)
+	if gerr != nil {
+		r.Log.Error(gerr, "ConfigVersions")
+		return ctrl.Result{}, err
```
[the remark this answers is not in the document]
dprince: Just pushed a fix. Thanks

PR #415, opened 2021-10-19
OSPK8-373 Render network_data.yaml from osnets

stuggi:
This allows rendering of network_data.yaml using the information of the osnets.
Also adds parameters for MTU and VIP to the osnet. With the VIP parameter
it can be controlled if a VIP should be requested for this network. E.g. there is
no need to have a VIP on the tenant network.


abays (commented):
Good stuff!  I just had one comment/question.

abays (approved):
/lgtm

PR #431, opened 2021-11-05
Use ENV variables to set Image URL defaults

dprince:
This is to better support installing our operator in offline mode.

OSPK8-394 #comment PR posted


abays (approved):
/lgtm

PR #435, opened 2021-11-11
Basic prototype of CR/CM/Secret backup/restore

abays:
Adds two new CRDs, `OpenStackBackupRequest` and `OpenStackBackup` to backup and restore CRs, ConfigMaps and Secrets.  Allows use of the first CRD to save/restore the second CRD.  This is the base mechanism, without any sort of cron operation.  If this will suffice for now, we can then move on to cron and backing-up other components.


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=osp-director-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.0+git
```
stuggi, line 9: is this an intended change 1.13.1 -> 1.13.0?

pkg/openstackbackup/funcs.go
```diff
@@ -0,0 +1,518 @@
...
+
+	if err != nil {
+		return *configMapList, err
+	}
+
+	// Also get certain config maps in other particular namespaces?
+	// TODO: which ones?
+
+	// Also get certain config maps used by our CRs
+	// TODO: which ones?
```
stuggi, line 129: we should backup the CMs/Secretes added via parameter to all CRs, like osplaybookgenerator: tarballConfigMap, gitSecret, heatEnvConfigMap

stuggi: or the controllers add backup label to the CM/Secret passed via a parameter so this can be used to filter instead of a named list

bundle.Dockerfile
abays: It's weird, but that is what it wants to put there.  I even have `1.13.1` installed:

```
# operator-sdk version
operator-sdk version: "v1.13.1", commit: "1659ab58a073999463e6dcdf18bdc359a315e091", kubernetes version: "1.21", go version: "go1.16.8", GOOS: "linux", GOARCH: "amd64"
```

I'm not sure why it generates `v1.13.0+git`.

pkg/openstackbackup/funcs.go
abays: We could look for existing `OsPlaybookGenerators` during backup and copy the CMs and Secrets.  I initially left them out since we do not save `OsPlaybookGenerators`.  I believe CMs and Secrets included as params in `OsPlaybookGenerators` don't have any ownership set on them, so I think it would be fine to save them without having the `OsPlaybookGenerators` in the backup.

To your second point: yes, we could also slap the controller label that I already look for on the CMs and Secrets when we processes `OsPlaybookGenerators`.

bundle.Dockerfile
stuggi: right, see this also in my latest patches. haven't realized before

pkg/openstackbackup/funcs.go
stuggi: I just thought it would be good to save the inputs for a current state we backup

bundle.Dockerfile
dprince: Is it because you installed from git that this is the case?

pkg/common/map.go
```diff
@@ -33,3 +33,18 @@ func MergeMaps(baseMap map[string]interface{}, extraMaps ...map[string]interface
 
 	return baseMap
 }
+
+// MergeStringMaps - merge two or more string->map maps
+func MergeStringMaps(baseMap map[string]string, extraMaps ...map[string]string) map[string]string {
+	if baseMap == nil {
+		baseMap = map[string]string{}
+	}
```
stuggi, line 41: could be `InitMap(&baseMap)`

api/v1beta1/openstackbackuprequest_types.go
```diff
@@ -0,0 +1,164 @@
...
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// OpenStackBackupRequestSpec defines the desired state of OpenStackBackupRequest
+type OpenStackBackupRequestSpec struct {
+	// Mode - what this backup should be doing (if it hasn't already done so)
+	// +kubebuilder:default=save
```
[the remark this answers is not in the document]
abays: Good call

controllers/openstackbackuprequest_controller.go
```diff
@@ -0,0 +1,569 @@
...
+func (r *OpenStackBackupRequestReconciler) setStatus(instance *ospdirectorv1beta1.OpenStackBackupRequest, oldStatus *ospdirectorv1beta1.OpenStackBackupRequestStatus, msg string) error {
+
+	if msg != "" {
+		r.Log.Info(msg)
+	}
+
+	if !reflect.DeepEqual(instance.Status, oldStatus) {
+		instance.Status.Conditions = ospdirectorv1beta1.ConditionList{}
+		// TODO: Using msg as reason and message for now
+		instance.Status.Conditions.Set(ospdirectorv1beta1.ConditionType(instance.Status.CurrentState), corev1.ConditionTrue, ospdirectorv1beta1.ConditionReason(msg), msg)
```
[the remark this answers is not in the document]
abays: Yes, I will try that now that we have your patch merged that added it.  Just did not have it at first.

main.go
```diff
@@ -270,6 +270,24 @@ func main() {
...
+	}
+
+	if err = (&controllers.OpenStackBackupRequestReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("controllers").WithName("OpenStackBackupRequest"),
+		Scheme: mgr.GetScheme(),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "OpenStackBackupRequest")
+		os.Exit(1)
+	}
```
abays, line 289: Need to inject Kclient here

pkg/common/map.go
abays: Ack

bundle.Dockerfile
abays: I did indeed follow the standard operator-SDK installation docs for doing so from Github: https://sdk.operatorframework.io/docs/installation/#install-from-github-release

stuggi:
good for me to go, if others are as well. In a follow up we probably want to integrate a kuttl test for b/r.

bundle.Dockerfile
dprince: Did I install from git source then? :)

controllers/openstackconfiggenerator.go
```diff
@@ -104,6 +104,18 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
...
 
+	// If we determine that a backup is overriding this reconcile, requeue after a longer delay
+	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, true)
+
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if overrideReconcile {
+		r.Log.Info(fmt.Sprintf("OpenStackClient %s reconcile overridden due to OpenStackBackupRequest(s) state; requeuing after 20 seconds", instance.Name))
```
stuggi, line 115: should this also be reflected in the conditions of each controller?

abays: Yes, that seems reasonable.  I will look into it after we get this PR merged.

stuggi (approved):
/lgtm

PR #437, opened 2021-11-12
Configure IPA client on openstackclient pod

olliewalsh:
Add secret to pass IPA config and admin creds to configure IPA client on
openstackclient pod.
Use kolla-config to persist files created/modified by IPA client install.

### DIFF
--- a/config/crd/bases/osp-director.openstack.org_openstackcontrolplanes.yaml
+++ b/config/crd/bases/osp-director.openstack.org_openstackcontrolplanes.yaml
@@ -83,6 +83,9 @@ spec:
                 description: GitSecret used to pull playbooks into the openstackclient
                   pod
                 type: string
+              idmSecret:
+                description: Idm secret used to register openstack client in IPA
+                type: string
               openStackClientImageURL:
                 description: OpenstackClient image
                 type: string
--- a/templates/openstackclient/bin/init.sh
+++ b/templates/openstackclient/bin/init.sh
@@ -24,15 +24,6 @@ fi
 # if the pvc is an empty volume, copy the existing hosts file to it
 if [ ! -f /mnt/etc/hosts ]; then
   cp /etc/hosts /mnt/etc/
-
-  # Ensure hostname -f and python socket.getfqdn() return the FQDN
-  if [ -v FQDN ]; then
-    SHORT_HOSTNAME=$(hostname -s)
-    LONG_HOSTNAME=$(hostname -f)
-    if [ "$LONG_HOSTNAME" != "$FQDN" ]; then
-      sed -i -e "s/^\([0-9.]\+\)\s\+\(.*\)${SHORT_HOSTNAME}\$/\1\t\2${FQDN} ${SHORT_HOSTNAME}/" /mnt/etc/hosts
-    fi
-  fi
 fi
 
 mkdir -p /home/cloud-admin/tripleo-deploy/validations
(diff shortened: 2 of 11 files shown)


templates/openstackclient/bin/init.sh
```diff
@@ -69,3 +60,59 @@ EOF
...
+  sudo bash -c "cat /mnt/etc/hostname > /etc/hostname"
+  # Ensure hostname -f and python socket.getfqdn() return the FQDN
+  SHORT_HOSTNAME=$(hostname -s)
+  LONG_HOSTNAME=$(hostname -f)
+  if [ "$LONG_HOSTNAME" != "$FQDN" ]; then
+    sed -i -e "s/^\([0-9.]\+\)\s\+\(.*\)${SHORT_HOSTNAME}\$/\1\t\2${FQDN} ${SHORT_HOSTNAME}/" /mnt/etc/hosts
+    sudo bash -c "cat /mnt/etc/hosts > /etc/hosts"
+  fi
+
+  cat <<EOF > /home/cloud-admin/openstackclient_ipa_install.yaml
```
olliewalsh, line 80: TODO: move this playbook to the tripleo-ipa repo.

It's based on the undercloud-ipa-install playbook but uses ipa-client-install directly instead of using the freeipa-ansible role (that fails as it tries to use hostnamectl which requires systemd).

templates/openstackclient/bin/init.sh
```diff
@@ -69,3 +60,59 @@ EOF
   git config --global user.email "dev@null.io"
   git config --global user.name "OSP Director Operator"
 fi
+
+
+# Skip the rest on tempest pod
+if cat /etc/hostname | grep ^tempest 2>&1 > /dev/null; then exit 0; fi
```
stuggi, line 66: instead of grep for tempest hostname, should we add a parameter to openstackclient and pass the info via a env var, so this part can be skipped on CR base when we create the tempest pod?

olliewalsh: Yes, this is just a quick and dirty hack to get tempest working. The tempest user isn't in sudoers so many of the init commands do not apply.

olliewalsh: Actually I think I can just drop that line - it was added before the next block was conditional on IPA_SERVER which will not be set for the tempest pod.

config/crd/bases/osp-director.openstack.org_openstackcontrolplanes.yaml
```diff
@@ -83,6 +83,9 @@ spec:
                 description: GitSecret used to pull playbooks into the openstackclient
                   pod
                 type: string
+              idmSecret:
```
stuggi, line 86: we should have a tls-e chapter in the Readme doc and explain how the secret content needs to look like, or have a separate doc file for tls-e?

olliewalsh: +1 will followup with docs for TLSe and TLS public endpoints

templates/openstackclient/bin/init.sh
olliewalsh: > Actually I think I can just drop that line - it was added before the next block was conditional on IPA_SERVER which will not be set for the tempest pod.

testing this now...

olliewalsh: works with that line removed

config/crd/bases/osp-director.openstack.org_openstackcontrolplanes.yaml
stuggi: sounds good thanks

PR #438, opened 2021-11-12
Convert CSV to image digests during build...

dprince:
Also, adds the missing kube-proxy to related images and finally adds the annotation to declare our operator supports offline mode.


abays (approved):
/lgtm

PR #439, opened 2021-11-15
Use lowercase role names to get VMs for fence CM generation

abays:
Role names for VMs belonging to `OsVMSets` for `OsControlPlanes` are ultimately stored as a lowercase label value in the child `VirtualMachineInstance` resources.  This might have changed in the past month or so, as we used to query against the capitalized role name for fencing data generation purposes (and it worked).  Anyhow, using lowercase role names fixes the problem and allows proper fencing config YAML generation.

### DIFF
--- a/controllers/openstackplaybookgenerator_controller.go
+++ b/controllers/openstackplaybookgenerator_controller.go
@@ -240,7 +240,7 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 			if common.StringInSlice(roleParams.RoleName, fencingRoles) && roleParams.RoleCount == 3 {
 				// Get the associated VM instances
 				virtualMachineInstanceList, err := common.GetVirtualMachineInstances(r, instance.Namespace, map[string]string{
-					common.OwnerNameLabelSelector: roleName,
+					common.OwnerNameLabelSelector: strings.ToLower(roleName),
 				})
 
 				if err != nil {


controllers/openstackplaybookgenerator_controller.go
```diff
@@ -240,7 +240,7 @@ func (r *OpenStackPlaybookGeneratorReconciler) Reconcile(ctx context.Context, re
 			if common.StringInSlice(roleParams.RoleName, fencingRoles) && roleParams.RoleCount == 3 {
 				// Get the associated VM instances
 				virtualMachineInstanceList, err := common.GetVirtualMachineInstances(r, instance.Namespace, map[string]string{
-					common.OwnerNameLabelSelector: roleName,
+					common.OwnerNameLabelSelector: strings.ToLower(roleName),
```
stuggi, line 243: I think the OwnerNameLabelSelector was always the object.Name [1] which afaik can only be lower case.
If we want the non lower ooo role name, we could add a label in [2] and use that.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/common/labels.go#L27
[2] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstackvmset_controller.go#L1068

stuggi (approved):
/lgtm

controllers/openstackplaybookgenerator_controller.go
abays: You're right.  In this case, though, the `OsVMSet` name is dictated by a RoleName from the `OsControlPlane`.  So, in turn, the `VirtualMachineInstance`s owner labels (the children of the `OsVMSet`) ultimately have the lowercase role name for `osp-director.openstack.org/name`.

PR #442, opened 2021-11-17
[OSPK8-328/OSPK8-388] Multi subnet support

stuggi:
Adds new CRDs openstacknetconfig and openstacknetattachment
* openstacknetconfig high level definition of all networks,
  subnets and network attachment

  Todo: cleanup networks/subnets if any got removed in a CR
        update.

* openstacknetattachment manages nncp and network-attachment-definitions
  specified in the osnetconfig upper level definition.
  Also manages the removal of bridge interfaces when an
  openstacknetattachment gets removed.

OSipset now renders the network_data.yaml to include the subnets
for both wallaby and train.

$TEMPLATES_DIR/environments/network-environment.yaml gets included
in the stack create command to pass the rendered network-environment
parameters for networks/subnets. Those parameters can be used
in the compute nic templates which allows to configure the routes
for the subnets.

Also:
* modifies the playbookgenerator to not not wait for osbmset
  if no host got requested. This allows to create osbmset with 0
  node count and still be able to render overcloud ansible playbooks.
* use defer function for osnetconfig, osnetatt, osnet to update the
  CR status. This has the benefit to only update the CR status when
  we return in the reconcile loop and reduces the number on possible
  concurrent writes to the status.
* adds an approach to have unified logging via common.LogForObject()
  and common.WrapErrorForObject()


api/v1beta1/openstacknetattachment_webhook.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Get the old bridge name, if any
+	var ok bool
+	var oldInstance *OpenStackNetAttachment
+
+	if oldInstance, ok = old.(*OpenStackNetAttachment); !ok {
+		return fmt.Errorf("runtime object is not an OpenStackNet")
```
abays, line 100: Hopefully this code is never reached, but I suppose `s/OpenStackNet/OpenStackNetAttachment/`

stuggi: thanks fixed with the latest added commit

controllers/openstacknet_controller.go
```diff
@@ -81,14 +71,14 @@ func (r *OpenStackNetReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=osp-director.openstack.org,resources=openstacknets/finalizers,verbs=update
 // +kubebuilder:rbac:groups=osp-director.openstack.org,namespace=openstack,resources=deployments/finalizers,verbs=update
 // FIXME: Cluster-scope required below for now, as the operator watches openshift-machine-api namespace as well
-// +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=create;delete;get;list;patch;update;watch
+// +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=create;delete;deletecollection;get;list;patch;update;watch
 // +kubebuilder:rbac:groups=nmstate.io,resources=nodenetworkconfigurationpolicies,verbs=create;delete;get;list;patch;update;watch
 
-// Reconcile -
+// Reconcile -, cond
```
abays, line 77: Looks like something slipped in here

controllers/openstacknet_controller.go
```diff
@@ -102,6 +92,52 @@ func (r *OpenStackNetReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+			if updateErr := r.Client.Status().Update(context.Background(), instance); updateErr != nil {
+				if err == nil {
+					err = common.WrapErrorForObject(
+						"Update Status", instance, updateErr)
+				} else {
+					common.LogErrorForObject(r, updateErr, "Update status", instance)
+				}
+			}
+		}
+	}(cond)
```
abays, line 139: Nice.  Is this a standard pattern for handling Conditions or did you create this yourself?

controllers/openstacknetattachment_controller.go
```diff
@@ -0,0 +1,675 @@
...
+	}
+
+	err = json.Unmarshal(desiredStateByte, &desiredState)
+	if err != nil {
+		cond.Message = fmt.Sprintf("Error unmarshal NodeNetworkConfigurationPolicy desired state: %v", string(desiredStateByte))
+		cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.NetAttachError)
+
+		err = common.WrapErrorForObject(cond.Message, instance, err)
+		return desiredState, err
+	}
```
abays, line 303: I know this works, but if we want to avoid the marshal/unmarshal approach, this might be an alternative: https://github.com/openstack-k8s-operators/osp-director-operator/blob/4c78bf791fb2b82255f21555f683e94c96684d23/pkg/nmstate/funcs.go#L9-L27.  Not important though.

controllers/openstacknetattachment_controller.go
```diff
@@ -0,0 +1,675 @@
...
+	if bridgeState != "absent" && bridgeState != "down" {
+		apply := func() error {
+			//
+			// Update nncp desired state to down to unconfigure the device on the worker nodes
+			//
+			re := regexp.MustCompile(`"state":"up"`)
+			desiredStateAbsent := re.ReplaceAllString(string(desiredState["desiredState"]), `"state":"down"`)
+			networkConfigurationPolicy.Spec.DesiredState = nmstateshared.State{
+				Raw: nmstateshared.RawState(desiredStateAbsent),
+			}
```
abays, line 496: Ah, will this fix the problem we work around [here](https://github.com/openstack-k8s-operators/osp-director-dev-tools/blob/455f27cb7ad5b5a61cb742fc6d182d0c393c2b7b/ansible/openstack_cleanup.yaml#L23-L31) in dev-tools?

controllers/openstackvmset_controller.go
```diff
@@ -281,22 +273,39 @@ func (r *OpenStackVMSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-		if osNetBindings[net] == "" {
+		// get network with name_lower label
+		network, err := openstacknet.GetOpenStackNetWithLabel(
+			r,
+			instance.Namespace,
+			map[string]string{
+				openstacknet.SubNetNameLabelSelector: netNameLower,
+			},
+		)
```
abays, line 288: I know you use this new function a lot.  Is it essentially a convenience function so you don't have to somehow convert the `NameLower` into the `Name` of the target resource?  I'm guessing that might be impossible anyhow.

pkg/openstackipset/configmap.go
```diff
@@ -24,13 +24,14 @@ import (
 	"strings"
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ospdirectorv1beta1 "github.com/openstack-k8s-operators/osp-director-operator/api/v1beta1"
 	common "github.com/openstack-k8s-operators/osp-director-operator/pkg/common"
 	"k8s.io/apimachinery/pkg/types"
 )
```
abays, line 32: I think the changes to this file would be worth going over in detail in the call tomorrow

pkg/openstacknetattachment/funcs.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, err
+	}
+	if len(osNetList.Items) == 0 {
+		return nil, k8s_errors.NewNotFound(v1.Resource("openstacknet"), fmt.Sprint(labelSelector))
+		//return nil, fmt.Errorf("OpenStackNet with label %v not found", labelSelector)
+	} else if len(osNetList.Items) > 1 {
+		return nil, fmt.Errorf("multiple OpenStackNet with label %v not found", labelSelector)
+	}
+	return &osNetList.Items[0], nil
+}
```
abays, line 109: Is this maybe a duplicate of the same func in `pkg/openstacknet/funcs.go`?

controllers/openstacknet_controller.go
stuggi: no, have not found this to be a standard pattern. in cnosp we used the defer to update the status, I enhanced this to manage the condition in that way

controllers/openstacknetattachment_controller.go
stuggi: nice, looks cleaner

controllers/openstacknetattachment_controller.go
stuggi: yes, this is the proper way to remove a previously configured device. need to remove that linked part in my dev-tools patch

controllers/openstackvmset_controller.go
stuggi: yes, because the osnet CR name must not be the NameLower name, e.g NameLower can have '_', while this is a forbidden char in CR names. That's why I add the ooo Name, NameLower, SubnetName as labels to the osnet CRs and then can use the ooo names to get the correct osnet.

pkg/openstacknetattachment/funcs.go
stuggi: yes, there are still things I need to cleanup

controllers/openstacknetattachment_controller.go
stuggi: changed/enhanced the nmstate funcs in last commit to use same approach

pkg/openstacknetattachment/funcs.go
stuggi: done now with commit https://github.com/openstack-k8s-operators/osp-director-operator/pull/442/commits/6a08932c41507f70674c7e05400e8b4e63ff2431  to use AttachType (AttachTypeBridge or AttachTypeSrIOV). extended osnetattch funcs.go and this osnet func to use those.

api/v1beta1/openstacknetattachment_types.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// AttachType -
+type AttachType string
+
+const (
+	// AttachTypeBridge -
+	AttachTypeBridge AttachType = "bridge"
+	// AttachTypeSrIOV -
+	AttachTypeSrIOV AttachType = "sriov"
```
olliewalsh, line 31: nit: can we use Sriov everywhere (or SRIOV)

dprince:
Nice work here. As the new CR is meant to be top level we should mark some of the other network CR's as "internal" objects:

https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/config/manifests/bases/osp-director-operator.clusterserviceversion.yaml#L10

api/v1beta1/openstacknetattachment_types.go
stuggi: donw

stuggi:
> Nice work here. As the new CR is meant to be top level we should mark some of the other network CR's as "internal" objects:
> 
> https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/config/manifests/bases/osp-director-operator.clusterserviceversion.yaml#L10

done, thanks for the pointer

tests/kuttl/tests/openstackbaremetalset_scale/03-assert.yaml
```diff
@@ -39,17 +40,19 @@ metadata:
...
   port: 6190
+  provisioningAgentImageUrl: quay.io/openstack-k8s-operators/provision-ip-discovery-agent@sha256:22db35fb3fb4bb2d67a29f6db46ca8cbd8a903cc4f3437cfaec11f2b6b6d3b6b 
 status:
   conditions:
   - message: ProvisionServer compute-provisionserver has been provisioned
     reason: ProvisionServer compute-provisionserver has been provisioned
     status: "True"
     type: Provisioned
+  localImageUrl: http://172.22.0.220:6190/images/rhel-guest-image-8.4-1168.x86_64.qcow2/compressed-rhel-guest-image-8.4-1168.x86_64.qcow2                                
+  provisionIp: 172.22.0.220
```
abays, line 55: Is the `172.22.0.220` IP here deterministic?  I just wonder if Metal3 dnsmasq might assign something different via DHCP.

stuggi: probably not, we could just remove checking those 2

PR #444, opened 2021-11-18
build script fixes: podman login check

dprince:
Corrects a bash error in the build script. Also runs a podman
login check to ensure users have permission to pull images
from the redhat.registry.io repo before running this script.

### DIFF
--- a/scripts/build_and_push_images.sh
+++ b/scripts/build_and_push_images.sh
@@ -7,6 +7,11 @@ set -ex
 # NOTE: Requires make, podman, opm and operator-sdk to be installed!
 #
 
+if ! podman login --get-login registry.redhat.io &> /dev/null; then
+  echo "Please run podman login registry.redhat.io before running this script."
+  exit 1
+fi
+
 VERSION=${1:-"17.0.1"}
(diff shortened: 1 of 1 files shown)


scripts/build_and_push_images.sh
```diff
@@ -38,7 +43,7 @@ sed -i 's/deploymentName: webhook/deploymentName: osp-director-operator-controll
 for csv_image in $(cat bundle/manifests/osp-director-operator.clusterserviceversion.yaml | grep "image:" | sed -e "s|.*image:||" | sort -u); do
   base_image=$(echo $csv_image | cut -f 1 -d':')
   tag_image=$(echo $csv_image | cut -f 2 -d':')
-  if ["$base_image:$tag_image" -eq "controller:latest"]; then
+  if [[ "$base_image:$tag_image" == "controller:latest" ]]; then
```
mpryc, line 46: double [[ ]] is extended test in bash for pattern matching between two strings, here one [ ]  would be sufficient, anyway no real change to the logic in this case.

mpryc (approved):
/lgtm

PR #450, opened 2021-11-30
OSPversion gathering/check

stuggi:
With the switch to the new way of setting defaults getting the OSP version from the provided client image was no longer used. This moves the OSP version check into the webhook where the default version is set when no specific OSP release info is set in the ctlplane CR.

Also changes the playbookgenerator script to use go template in create-playbooks.sh for version conditions instead of bash if.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/451


abays (approved):
/lgtm

abays (approved):
/lgtm

PR #451, opened 2021-11-30
[OSPK8-324 OSPK8-325] DeployedNetworkEnvironment and CtlplaneNetworkAttributes

stuggi:
Renders DeployedNetworkEnvironment and CtlplaneNetworkAttributes from network information required by wallaby/OSP17 networkv2.

New Parameters:
 - IsControlPlane to make it easier to identify controlplane network which is kind of special network from ooo side. The webhook sets the parameter automatically for the default ctlplane names when Name is "Control" or NameLower "ctlplane".
  Parameters required to render DeployedNetworkEnvironment and CtlplaneNetworkAttributes:
 - DomainName the name of the dns domain, this is used to create the per ooo network domain name, like `internalapi.osptest.test.metalkube.org`
 - DNSServers list of dns servers
 - DNSSearchDomains list of DNS search domains

TODO - in a follow up we probably want to:
* change osctlplane and osbms to get the dns servers from the osnetcfg.
* add DnsServers + DnsSearchDomain in parameter_defaults of the rendered 
* set CloudName and CloudName<NetworkName> fro the provided DomainName

Depends-On: https://github.com/openstack-k8s-operators/osp-director-operator/pull/452

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -57,8 +57,8 @@ resources:
   path: github.com/openstack-k8s-operators/osp-director-operator/api/v1beta1
   version: v1beta1
   webhooks:
-    validation: true
     defaulting: true
+    validation: true
     webhookVersion: v1
 - controller: true
   domain: openstack.org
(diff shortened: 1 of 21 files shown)


abays (approved):
/lgtm

PR #452, opened 2021-11-30
[OSPK8-402] Add bridge name validation to osnetcfg webhook

stuggi:
This adds validation check to the osnetcfg webhook to prevent bridge name changes on an update.


abays (approved):
/lgtm

PR #457, opened 2021-12-03
[OSPK8-405 OSPK8-407] osipset/osnet/osplaybookgenerator rework

stuggi:
* Rework osipset to use defer function to update instance status 
Also makes SortMapByValue generic and add to map.go

* Rework osipset/osnet to store reservations in osnet spec 
Also fixes an issue with getNormalizedStatus() which resulted in not
setting the condition time stamps to an empty Time for the returning
object.

* Rework osplaybookgenerator to use defer function to update instance status
* Integrate heat env rendering into osplaybookgenerator also
Moves the rendering onf heat environment from osipset to the
osplaybookgenerator.

Also optimize the CMs the osplaybookgenerator watches by adding a PlaybookGeneratorInputLabel
to the customer provided configmaps to limit the events to reconcile to only the really required data
which the CR does not own.


pkg/common/map.go
```diff
@@ -33,3 +35,45 @@ func MergeMaps(baseMap map[string]interface{}, extraMaps ...map[string]interface
...
 	return baseMap
 }
+
+// MergeStringMaps - merge two or more string->map maps
+func MergeStringMaps(baseMap map[string]string, extraMaps ...map[string]string) map[string]string {
+	InitMap(&baseMap)
+
+	for _, extraMap := range extraMaps {
+		for key, value := range extraMap {
+			baseMap[key] = value
```
bogdando, line 45: this looks like Joining (with overriding existing keys), not merging?

stuggi: it merges/joins, but yes, if the key already exists, it will be overwritten

controllers/openstackconfiggenerator.go
```diff
@@ -630,3 +651,335 @@ func (r *OpenStackConfigGeneratorReconciler) verifyNodeResourceStatus(instance *
...
+		{
+			Name:               "tripleo-deploy-config",
+			Namespace:          instance.Namespace,
+			Type:               common.TemplateTypeConfig,
+			InstanceType:       instance.Kind,
+			AdditionalTemplate: map[string]string{},
+			CustomData: common.MergeStringMaps(
+				roleNicTemplates,
+				fencingTemplate,
+			),
```
abays, line 833: Just curious as to why `common.MergeStringMaps` is called on 818 and 830 with the same parameters.

stuggi: Line 818 is nonsense, can not remember why it is there. probably I wanted to use a var, but for whatever reason stopped that. We can just remove 818.

controllers/openstackipset_controller.go
```diff
@@ -485,37 +449,47 @@ func (r *OpenStackIPSetReconciler) addIP(instance *ospdirectorv1beta1.OpenStackI
...
+					instance,
+				)
 			}
 
 			// add netName -> reservationIP to hostIPs map
 			hostNetworkIPs[netName] = reservationIP
 
-			err = r.Client.Status().Update(context.TODO(), network)
+			// update status of OSNet
+			// TODO: just patch?
```
abays, line 487: I know you put this `TODO` in a few places, but I just had a thought on it.  It seems that either would be fine here, but it's easier, I think, to just do the `Update` if you already have the whole object.  Maybe `Patch` can dodge the `"objected has been modified"` error in some cases though?

stuggi: yes, I agree that for now we should be good to just update the full object. I was more thinking on when used server side apply there is a field management mechanism and if a different manager updates a field it not owns it will result in a conflict. if we get to this and see it, we might need to use patch to just update a single field.

controllers/openstackconfiggenerator.go
stuggi: removed in new commit

controllers/openstackipset_controller.go
abays: Yeah, I think it would be interesting to investigate using `Patch` here (but not right now) and elsewhere in an effort to reduce the occurrences of conflicts

abays (approved):
Seems ready to me.  Let's merge if no one else wants to review and it passes AI CI.

PR #460, opened 2021-12-08
[OSPK8-416] reduce repeated condition message logging

stuggi:
Right now on every reconcile messages get logged, even if there
was no action.

To make logging better readable/ easier to follow, this change
logs the condition messages in the defer function where we check if
the condition changed. This should be used where we return from the
reconcile loop, and only in cases where this is possible log explicitely
at that place.

### DIFF
--- a/controllers/openstacknetconfig_controller.go
+++ b/controllers/openstacknetconfig_controller.go
@@ -127,6 +127,9 @@ func (r *OpenStackNetConfigReconciler) Reconcile(ctx context.Context, req ctrl.R
 				} else {
 					common.LogErrorForObject(r, updateErr, "Update status", instance)
 				}
+			} else {
+				// log status changed messages also to operator log
+				common.LogForObject(r, cond.Message, instance)
 			}
 		}
 	}(cond)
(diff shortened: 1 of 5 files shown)


controllers/openstacknetconfig_controller.go
```diff
@@ -253,7 +256,7 @@ func (r *OpenStackNetConfigReconciler) Reconcile(ctx context.Context, req ctrl.R
 		} else if !reflect.DeepEqual(ctrlResult, ctrl.Result{}) {
 			cond.Message = fmt.Sprintf("OpenStackNetConfig %s waiting for all OpenStackNetworkAttachments to be configured", instance.Name)
 			cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.NetConfigWaiting)
-			common.LogForObject(r, cond.Message, instance)
+			//common.LogForObject(r, cond.Message, instance)
```
abays, line 259: Remove this line entirely?

controllers/openstacknetconfig_controller.go
```diff
@@ -292,7 +295,7 @@ func (r *OpenStackNetConfigReconciler) Reconcile(ctx context.Context, req ctrl.R
 			} else if !reflect.DeepEqual(ctrlResult, ctrl.Result{}) {
 				cond.Message = fmt.Sprintf("OpenStackNetConfig %s waiting for all OpenStackNetworks to be configured", instance.Name)
 				cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.NetConfigWaiting)
-				common.LogForObject(r, cond.Message, instance)
+				//common.LogForObject(r, cond.Message, instance)
```
abays, line 298: Remove this line entirely?

controllers/openstacknetconfig_controller.go
stuggi: indeed leftover from testing...

stuggi: done in force pushed commit

controllers/openstacknetconfig_controller.go
stuggi: same

abays (approved):
/lgtm

PR #466, opened 2021-12-13
[OSPK8-412] convert openstackclient to defer

stuggi:


pkg/common/secret.go
```diff
@@ -282,3 +286,55 @@ func DeleteSecretsWithLabel(r ReconcilerCommon, obj metav1.Object, labelSelector
...
+		cond.Message = fmt.Sprintf("Error getting %s Secret: %v", secretName, err)
+		cond.Reason = ospdirectorv1beta1.ConditionReason(conditionDetails.ConditionErrordReason)
+		cond.Type = ospdirectorv1beta1.ConditionType(conditionDetails.ConditionErrorType)
+		err = WrapErrorForObject(cond.Message, object, err)
+
+		return data, ctrl.Result{}, err
+	}
+
+	if key != "" {
+		if _, ok := secret.Data[key]; !ok {
```
bogdando, line 326: just a minor nit, but you could as well save the key for an externally (to if) scoped variable for future use at the line #336 and save an extra Data[key] accessing off

stuggi: yes correct, let me change that

controllers/openstackclient_controller.go
```diff
@@ -385,32 +399,48 @@ func (r *OpenStackClientReconciler) podCreateOrUpdate(instance *ospdirectorv1bet
...
 		)
 		if err != nil {
 			if k8s_errors.IsNotFound(err) {
-				r.Log.Info(fmt.Sprintf("OpenStackNet with NameLower %s not found!", netName))
+				common.LogForObject(
+					r,
+					fmt.Sprintf("OpenStackNet with NameLower %s not found!", netNameLower),
+					instance,
+				)
 				continue
```
abays, line 422: Should we make this similar to lines 652-659 in terms of logging the error and updating the status?

controllers/openstackclient_controller.go
```diff
@@ -546,22 +576,272 @@ func (r *OpenStackClientReconciler) podCreateOrUpdate(instance *ospdirectorv1bet
...
+			// e.g. additional controller VM got up. We just re-create the
+			// openstackclient pod
+			common.LogForObject(
+				r,
+				fmt.Sprintf("openstackclient pod deleted due to spec change %v", err),
+				instance,
+			)
+			if err := r.Client.Delete(context.TODO(), pod); err != nil {
+
+				return err
```
abays, line 601: Same question here re: the error.  Any need to log or store it in status?

controllers/openstackclient_controller.go
stuggi: yes we probably should do that, but we should never see this error here as we already created the ipset where the networks is also checked, but lets add it and not continue, instead return a real error with condition.

controllers/openstackclient_controller.go
stuggi: yes we should do that, but ignore k8s_errors.IsNotFound

pkg/openstackipset/funcs.go
```diff
@@ -56,3 +65,69 @@ func OvercloudipsetCreateOrUpdate(r common.ReconcilerCommon, obj metav1.Object,
...
+		ipSet.Spec.RoleName = ipsetDetails.Role
+		ipSet.Spec.HostCount = ipsetDetails.HostCount
+		ipSet.Spec.VIP = ipsetDetails.VIP
+		ipSet.Spec.AddToPredictableIPs = ipsetDetails.AddToPredictableIPs
+		ipSet.Spec.HostNameRefs = ipsetDetails.HostNameRefs
+
+		err := controllerutil.SetControllerReference(object, ipSet, r.GetScheme())
+
+		if err != nil {
+			return err
```
abays, line 100: Also same question here I suppose

stuggi: indeed we should

stuggi (commented):
thanks for the pointers, I'll update them

abays (approved):
/lgtm

PR #467, opened 2021-12-13
Fix typo where TIP is master not main

mpryc:
Small typo which results in building containers that are
tagged with master-latest rather then latest for the master
branch. For other branches it's $branch-latest.


stuggi (approved):
/lgtm

PR #469, opened 2021-12-14
Add config map to install custom CA certs

olliewalsh:
Required for TLS public endpoints where we cannot rely on IPA client to
install CA certs.

### DIFF
--- a/pkg/common/configmap.go
+++ b/pkg/common/configmap.go
@@ -19,11 +19,14 @@ package common
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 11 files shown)


pkg/common/configmap.go
```diff
@@ -202,3 +204,39 @@ func GetConfigMapAndHashWithName(r ReconcilerCommon, configMapName string, names
...
+	err := r.GetClient().Get(context.TODO(), types.NamespacedName{Name: configMapName, Namespace: object.GetNamespace()}, configMap)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			cond.Message = fmt.Sprintf("%s config map does not exist: %v", configMapName, err)
+			cond.Reason = ospdirectorv1beta1.ConditionReason(conditionDetails.ConditionNotFoundReason)
+			cond.Type = ospdirectorv1beta1.ConditionType(conditionDetails.ConditionNotFoundType)
+
+			LogForObject(r, cond.Message, object)
+
+			return configMap, ctrl.Result{}, err
```
stuggi, line 231: From description we want to reconcile when not found, so should be:

`return configMap, ctrl.Result{RequeueAfter: time.Duration(requeueTimeout) * time.Second}, nil`

stuggi (commented):
looks good, just one small change needed as the return from GetConfigMap is missing the timeout to reconcile.

pkg/common/configmap.go
olliewalsh: yes, well spotted!

stuggi (approved):
/lgtm

PR #464, opened 2021-12-10
Updates to avoid CVE's

dprince:

PR #473, opened 2021-12-15
Updates to avoid CVE's (#464)

dprince:
* go.mod updates for security and CVEs

* Bump CDI and github.com/ulikunitz

* Drop prometheus/common dependency

This avoids using the logger from prometheus/common and
instead uses the controller-runtime setup logger instead

* Bump miekg/dns

* Correct CVE comment for containerd


stuggi (approved):
/lgtm

PR #475, opened 2021-12-15
Log in to redhat.io registry for bundle generation

mpryc:
Bundle requires login to the redhat.io registry to properly
calculate hashes.


stuggi (approved):
/lgtm

PR #476, opened 2021-12-15
Fix controller VM DNS config

olliewalsh:
/etc/resolv.conf on the controller VMs is still referencing the OCP
cluster DNS server on 172.30.0.10, which is not responding.

This is resulting in slow DNS queries as it wait to timeout before
trying the next/correct server in resolv.conf.

This disables the default DNS config when BootstrapDNS is set, similar
to the baremetal node.


stuggi (approved):
/lgtm

PR #478, opened 2021-12-20
[OSPK8-314] Openstack deploy

dprince:


containers/agent/deploy.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return err
+	}
+
+	argString := fmt.Sprintf("%s\n", command)
+	reader := strings.NewReader(argString)
+
+	return exec.Stream(remotecommand.StreamOptions{
+		Stdin:  io.Reader(reader),
+		Stdout: os.Stdout,
```
dprince, line 79: @abays I modified these to use osStdout and os.Stderr. That way we get all the data streamed back to the pod logs in real time instead of once at the end.

config/default/manager_default_images.yaml
```diff
@@ -25,6 +25,8 @@ spec:
           value: registry.redhat.io/rhosp-rhel8-tech-preview/osp-director-downloader:1.0
         - name: PROVISIONING_AGENT_IMAGE_URL_DEFAULT
           value: registry.redhat.io/rhosp-rhel8-tech-preview/osp-director-provisioner:1.0
+        - name: AGENT_IMAGE_URL_DEFAULT
+          value: quay.io/dprince/osp-director-operator-agent:0.0.1
```
stuggi, line 29: quay.io/dprince/osp-director-operator-agent should be quay.io/openstack-k8s-operators/osp-director-operator-agent

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/osp-director-operator
-  newTag: 17.0.1
+  newName: quay.io/dprince/osp-director-operator
+  newTag: 0.0.66
```
stuggi, line 8: should not change this file

config/samples/osp-director_v1beta1_openstackdeploy.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: osp-director.openstack.org/v1beta1
+kind: OpenStackDeploy
+metadata:
+  name: openstackdeploy-sample
+spec:
+  configVersion: bar
+  configGenerator: foo
```
stuggi, line 7: should list optional imageURL param?

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// DeployGetLabels -
+func DeployGetLabels(name string) map[string]string {
+	return map[string]string{"owner": "osp-director-operator", "cr": name, "app": "deploy"}
+}
```
stuggi, line 219: looks like this is not being used?

containers/agent/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	componentName = "osp-director-operator-agent"
+)
+
+var (
+	rootCmd = &cobra.Command{
+		Use:   componentName,
+		Short: "Run OSP-Director-Operator Deploy Agent",
+		Long:  "Runs the OSP-Director-Operator Deploy Agent",
```
abays, line 18: Do we want to make this more generic now, i.e. remove `Deploy`, in anticipation of reusing this agent for more than just the deploy option?

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// look up the git secret from the configGenerator
+	configGenerator := &ospdirectorv1beta1.OpenStackConfigGenerator{}
+	err = r.GetClient().Get(context.TODO(), types.NamespacedName{Name: instance.Spec.ConfigGenerator, Namespace: instance.Namespace}, configGenerator)
+	if err != nil && errors.IsNotFound(err) {
+		r.GetLogger().Error(err, instance.Spec.ConfigGenerator+" ConfigGenerator not found!", "Instance.Namespace", instance.Namespace)
+	}
+
+	// Define a new Job object
+	job := openstackdeploy.DeployJob(instance, "openstackclient", instance.Spec.ConfigVersion, configGenerator.Spec.GitSecret)
```
abays, line 146: Is this used outside the scope of the if-block that starts on 148?  If not, we could move it inside to avoid calling the `DeployJob` func when the `ConfigVersion` matches between `Status` and `Spec`?

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+		}
+
+	}(cond)
+
+	// look up the git secret from the configGenerator
+	configGenerator := &ospdirectorv1beta1.OpenStackConfigGenerator{}
+	err = r.GetClient().Get(context.TODO(), types.NamespacedName{Name: instance.Spec.ConfigGenerator, Namespace: instance.Namespace}, configGenerator)
+	if err != nil && errors.IsNotFound(err) {
+		r.GetLogger().Error(err, instance.Spec.ConfigGenerator+" ConfigGenerator not found!", "Instance.Namespace", instance.Namespace)
```
abays, line 142: Could consider using `common.WrapErrorForObject` here

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,241 @@
...
+				common.LogForObject(r, cond.Message, instance)
+			}
+		}
+
+	}(cond)
+
+	// look up the git secret from the configGenerator
+	configGenerator := &ospdirectorv1beta1.OpenStackConfigGenerator{}
+	err = r.GetClient().Get(context.TODO(), types.NamespacedName{Name: instance.Spec.ConfigGenerator, Namespace: instance.Namespace}, configGenerator)
+	if err != nil && errors.IsNotFound(err) {
```
abays, line 141: Should we also check the state of the `ConfigGenerator` to make sure it's done running, or do we account for that in some other way?

config/default/manager_default_images.yaml
dprince: agreed. Before I did that however I wanted to get feedback on the container name in general. Sounds like it is a go.

containers/agent/main.go
dprince: good call. will do

controllers/openstackdeploy_controller.go
dprince: good call, will move inside the if-block

controllers/openstackdeploy_controller.go
dprince: The ConfigGenerator is just being referenced here to lookup the associated GitSecret so I don't think it matters if it is actually running. The OpenStackDeploy CR's use ConfigVersions (created by the ConfigGenerator when it finishes running an iteration) so by design there should be no problem if say the ConfigGenerator runs a 2nd time because of some new configuration change. A change like this would result in a new ConfigVersion which wouldn't get deployed until you created an OpenStackDeploy CR referencing it.

config/manager/kustomization.yaml
dprince: Fixed in latest commits

config/samples/osp-director_v1beta1_openstackdeploy.yaml
dprince: well, I suppose I'd like to have the example just rely on our default. I'd actually like to go the other way and have most examples remove the imageURLs

controllers/openstackdeploy_controller.go
dprince: thanks, removed in latest set of commits

containers/agent/main.go
dprince: fixed in latest commits

controllers/openstackdeploy_controller.go
dprince: done in the latest commits

abays (commented):
I'm good with merging this if we think it's ready

config/samples/osp-director_v1beta1_openstackdeploy.yaml
stuggi: ok, maybe we should have more samples. Non defaults in a subdir, like for kuttl tests where we use/validate some more parameters not set per default

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		op, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, job, func() error {
+			err := controllerutil.SetControllerReference(instance, job, r.Scheme)
+			if err != nil {
+				return err
+			}
+
+			return nil
+		})
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
```
stuggi, line 158: could set a generic job error condition here

controllers/openstackdeploy_controller.go
```diff
@@ -0,0 +1,235 @@
...
+				if err == nil {
+					err = common.WrapErrorForObject(
+						"Update Status", instance, updateErr)
+				} else {
+					common.LogErrorForObject(r, updateErr, "Update status", instance)
+				}
+			} else {
+				// log status changed messages also to operator log
+				common.LogForObject(r, cond.Message, instance)
+			}
```
stuggi, line 133: for the other controllers we switched to always log the condition to the operator log, even if it has not changed so that its better to follow the reconciliations from the operator log [1]

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstacknetconfig_controller.go#L128-L139

pkg/openstackdeploy/const.go
```diff
@@ -0,0 +1,29 @@
...
+
+package openstackdeploy
+
+const (
+	// AppLabel -
+	//AppLabel = "osp-openstackconfiggenerator"
+
+	// ConfigGeneratorInputLabel - label set on objects which are input for the configgenerator
+	// but not owned by the configgenerator
+	//ConfigGeneratorInputLabel = "playbook-generator-input"
```
stuggi, line 25: remove or update to Deploy?

pkg/openstackdeploy/const.go
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstackdeploy
+
+const (
+	// AppLabel -
+	//AppLabel = "osp-openstackconfiggenerator"
```
stuggi, line 21: want to remove if not used?

config/manager/kustomization.yaml
olliewalsh: should we add this to .gitignore?

controllers/openstackdeploy_controller.go
dprince: ack, I will make the switch.

api/v1beta1/openstackdeploy_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// CurrentReason
+	CurrentReason DeployReason `json:"currentReason"`
+
+	Conditions ConditionList `json:"conditions,omitempty" optional:"true"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:shortName=osdeploy;osdeploys
```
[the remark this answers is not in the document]
dprince: ack, will add it

api/v1beta1/webhook_suite_test.go
```diff
@@ -72,6 +72,9 @@ var _ = BeforeSuite(func() {
 	err = admissionv1beta1.AddToScheme(scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+err = admissionv1beta1.AddToScheme(scheme	)
+Expect(err).NotTo(HaveOccurred())
+
```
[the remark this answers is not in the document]
dprince: ack

api/v1beta1/openstackdeploy_types.go
```diff
@@ -0,0 +1,109 @@
...
+//+operator-sdk:csv:customresourcedefinitions:displayName="OpenStack Deploy"
+//+kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.currentState`,description="Status"
```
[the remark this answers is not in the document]
dprince: okay, do we want to do this all in one patch at a later time and leave these in for now?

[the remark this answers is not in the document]
stuggi: I think we can leave this for now and do it in set of patches for all controllers

stuggi (approved):
/lgtm

api/v1beta1/webhook_suite_test.go
```diff
@@ -93,6 +93,9 @@ var _ = BeforeSuite(func() {
 	err = (&OpenStackNetAttachment{}).SetupWebhookWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
 
+err = (&OpenStackDeploy{}).SetupWebhookWithManager(mgr)
```
olliewalsh, line 96: nit: indent

main.go
```diff
@@ -288,6 +288,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controllers.OpenStackDeployReconciler{
+		Client:  mgr.GetClient(),
+		Kclient: kclient,
+		Log:     ctrl.Log.WithName("controllers").WithName("OpenStackBackupRequest"),
```
olliewalsh, line 294: wrong controller name

dprince: cool. Fixed in the latest commit.

api/v1beta1/webhook_suite_test.go
dprince: This isn't needed. I'll just remove in the latest patch.

PR #479, opened 2021-12-21
[OSPK8-411] Migrate OSCtlPlane to use defer

stuggi:


controllers/openstackcontrolplane_controller.go
```diff
@@ -695,20 +591,454 @@ func (r *OpenStackControlPlaneReconciler) createOrUpdateOpenStackMACAddress(inst
...
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.MACCondReasonCreateMACError)
+		cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.CommonCondTypeError)
+		err = common.WrapErrorForObject(cond.Message, instance, err)
+
 		return err
 	}
 	if op != controllerutil.OperationResultNone {
-		r.Log.Info(fmt.Sprintf("OpenStackMACAddress CR successfully reconciled - operation: %s", string(op)))
+		cond.Message = fmt.Sprintf("OpenStackMACAddress CR successfully reconciled - operation: %s", string(op))
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.OsClientCondReasonPodProvisioned)
```
abays, line 614: Is this the right condition reason?

controllers/openstackcontrolplane_controller.go
```diff
@@ -695,20 +591,454 @@ func (r *OpenStackControlPlaneReconciler) createOrUpdateOpenStackMACAddress(inst
...
+			err = common.EnsureSecrets(r, instance, pwSecret, envVars)
+			if err != nil {
+				cond.Message = fmt.Sprintf("Error creating TripleoPasswordsSecret %s", controlplane.TripleoPasswordSecret)
+				cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.ControlPlaneReasonTripleoPasswordsSecretCreateError)
+				cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.CommonCondTypeError)
+				err = common.WrapErrorForObject(cond.Message, instance, err)
+
+				return err
+			}
+		} else if err != nil {
```
abays, line 667: Could this just be an `else`, since we already know `err` is non-nil here?

controllers/openstackcontrolplane_controller.go
```diff
@@ -695,20 +591,454 @@ func (r *OpenStackControlPlaneReconciler) createOrUpdateOpenStackMACAddress(inst
...
+		common.LogForObject(
+			r,
+			fmt.Sprintf("Updating CR status with new hostname information, %d new - %s",
+				len(newVMs),
+				diff.ObjectReflectDiff(currentNetStatus, instance.Status.VIPStatus),
+			),
+			instance,
+		)
+
+		err := r.Client.Status().Update(context.TODO(), instance)
```
abays, line 883: Would the defer in the main reconcile loop not try to update the status anyhow, since it would compare `Status` to the old deep copy and see the difference(s)?

controllers/openstackcontrolplane_controller.go
```diff
@@ -695,20 +591,454 @@ func (r *OpenStackControlPlaneReconciler) createOrUpdateOpenStackMACAddress(inst
...
+			cond.Message = fmt.Sprintf("Failed to create or update %s %s ", vmSet.Kind, vmSet.Name)
+			cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.VMSetCondReasonError)
+			cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.CommonCondTypeError)
+			err = common.WrapErrorForObject(cond.Message, instance, err)
+
+			return vmSets, err
+		}
+		if op != controllerutil.OperationResultNone {
+			cond.Message = fmt.Sprintf("%s %s operation: %s", vmSet.Kind, vmSet.Name, string(op))
+			cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.VMSetCondReasonProvisioned)
```
abays, line 966: Isn't the VMSet just created/updated here, and not necessarily provisioned yet?

controllers/openstackcontrolplane_controller.go
stuggi: yes, good catch

controllers/openstackcontrolplane_controller.go
stuggi: indeed

controllers/openstackcontrolplane_controller.go
stuggi: yes it will when we return. since the hostname is some important information to create the IPs for I thought it might make sense to update it right after created them and fail/reconcile early if it fail for whatever reasone.

controllers/openstackcontrolplane_controller.go
stuggi: yes correct, will update that here

controllers/openstackcontrolplane_controller.go
abays: Ack, makes sense

abays (approved):
/lgtm

PR #480, opened 2021-12-21
Add param to include t-h-t environment files

olliewalsh:
Add HeatEnv param to include built-in t-h-t environment files in stack
create command line.


api/v1beta1/openstackconfiggenerator_types.go
```diff
@@ -24,6 +24,8 @@ import (
 type OpenStackConfigGeneratorSpec struct {
 	// Name of the image used to generate configs. If missing will be set to the configured OPENSTACKCLIENT_IMAGE_URL_DEFAULT in the CSV for the OSP Director Operator.
 	ImageURL string `json:"imageURL,omitempty"`
+	// Optional. List of heat environment files to include from tripleo-heat-templates/environments
+	HeatEnvs []string `json:"heatEnvs,omitempty"`
```
stuggi, line 28: maybe default to an empty list like here https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/api/v1beta1/openstacknetattachment_types.go#L37

controllers/openstackconfiggenerator.go
```diff
@@ -263,6 +264,18 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
+	// add build-in heat environment files
+	var tripleoEnvironmentFiles []string
+	if instance.Spec.HeatEnvs != nil {
```
stuggi, line 269: with the above set default to an empty slice I think we won't require this.

config/crd/bases/osp-director.openstack.org_openstackconfiggenerators.yaml
```diff
@@ -89,6 +89,11 @@ spec:
                 description: Required. the name of the config map containing Heat
                   env file customizations
                 type: string
+              heatEnvs:
+                description: List of heat environment files to include from tripleo-heat-templates/environments
```
stuggi, line 93: should we be here really specific that it is just the file name without any path?

controllers/openstackconfiggenerator.go
olliewalsh: yes, avoiding the nil check is why I wanted to set a default

api/v1beta1/openstackconfiggenerator_types.go
olliewalsh: will try it now...

config/crd/bases/osp-director.openstack.org_openstackconfiggenerators.yaml
olliewalsh: can include a path, but it's relative to t-h-t/environments

olliewalsh: e.g ssl/enable-internal-tls.j2.yaml

stuggi: oh right, sure. yes we need subnet references.

api/v1beta1/openstackconfiggenerator_types.go
```diff
@@ -24,6 +24,10 @@ import (
 type OpenStackConfigGeneratorSpec struct {
 	// Name of the image used to generate configs. If missing will be set to the configured OPENSTACKCLIENT_IMAGE_URL_DEFAULT in the CSV for the OSP Director Operator.
 	ImageURL string `json:"imageURL,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={}
```
stuggi, line 28: well I am not sure if that really works as I was expecting it to do. we do not see a default in the rendered crd yaml. maybe we should default it in the webhook if needed.

olliewalsh: Does the right thing AFAICT. I deployed with and without this set

olliewalsh: No default in the crd yaml when it's optional maybe?

dprince (approved):

PR #481, opened 2022-01-04
[OSPK8-413] convert osbmset to defer

stuggi:
Converts the osbmset controller to use defer function to update/set conditions.

Also:
* Allign defer function of all controllers to log current cond msg
  The condition gets updated only if the condition of the CR
  changed, but we should log all of them to be able to follow the
  reconcilations from the operator log.

* Set condition if finalizer fails to be removed

### DIFF
--- a/api/v1beta1/openstackbaremetalset_types.go
+++ b/api/v1beta1/openstackbaremetalset_types.go
@@ -136,17 +136,6 @@ const (
 	BaremetalSetCondReasonVirtualMachineProvisioned ConditionReason = "BaremetalHostProvisioed"
 	// BaremetalSetCondReasonVirtualMachineCountZero - no bmh requested
 	BaremetalSetCondReasonVirtualMachineCountZero ConditionReason = "BaremetalHostCountZero"
-
-	/*
-		// BaremetalSetReasonListError - osbms list objects error
-		BaremetalSetReasonListError ConditionReason = "BaremetalSetReasonListError"
-		// BaremetalSetReasonNotFound - osbms object not found
-		BaremetalSetReasonNotFound ConditionReason = "BaremetalSetReasonNotFound"
-		// BaremetalSetReasonNotReady - osbms not yet ready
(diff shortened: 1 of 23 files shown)


api/v1beta1/openstackbaremetalset_types.go
```diff
@@ -80,20 +80,62 @@ type OpenStackBaremetalSetProvisioningStatus struct {
...
+	// BaremetalSetCondReasonUserDataSecretDeleteError - error deleting user data secret
+	BaremetalSetCondReasonUserDataSecretDeleteError ConditionReason = "BaremetalSetUserDataSecretDeleteError"
+	// BaremetalSetCondReasonNetworkDataSecretDeleteError - error deleting network data secret
+	BaremetalSetCondReasonNetworkDataSecretDeleteError ConditionReason = "BaremetalSetNetworkDataSecretDeleteError"
+	// BaremetalSetCondReasonScaleDownInsufficientAnnotatedHosts - not enough nodes annotated for deletion
+	BaremetalSetCondReasonScaleDownInsufficientAnnotatedHosts ConditionReason = "BaremetalSetScaleDownInsufficientAnnotatedHosts"
+	// BaremetalSetCondReasonScaleUpInsufficientHosts - not enough nodes for requested host count
+	BaremetalSetCondReasonScaleUpInsufficientHosts ConditionReason = "BaremetalSetScaleUpInsufficientHosts"
+	// BaremetalSetCondReasonProvisioningErrors - errors during bmh provisioning
+	BaremetalSetCondReasonProvisioningErrors ConditionReason = "BaremetalSetProvisioningErrors"
```
abays, line 130: Should we make the value (but not the variable name) `BaremetalHostProvisioningErrors` to match with the other BMH-related condition reasons that follow?  If this is how you want it though, no worries

abays: Nevermind.  I just saw how you use it in the overall status calculation for the BMSet controller.  I think it makes sense how you have it now.

abays (approved):
/lgtm

PR #482, opened 2022-01-05
[OSPK8-422] generate ctlplane Ansible inventory

dprince:
This patch creates an Ansible inventory of the cltplane
IPs within the openstackclient pod. The intent is for this
to be used for RH node registration instead of relying
on the ConfigGenerator/TripleO inventory to be generated
which doesn't actually exist at the time we need it with
the new workflow.

### DIFF
--- a/pkg/openstacknet/funcs.go
+++ b/pkg/openstacknet/funcs.go
@@ -85,3 +85,25 @@ func GetOpenStackNetWithLabel(r common.ReconcilerCommon, namespace string, label
 	}
 	return &osNetList.Items[0], nil
 }
+
+// GetOpenStackNetworkRoleHostnameIPs - Return a map of hostname/IPs for the specified network selector
+func GetOpenStackNetworkRoleHostnameIPs(r common.ReconcilerCommon, namespace string, labelSelector map[string]string) (map[string]map[string]string, error) {
+	ctlplaneIps := map[string]map[string]string{}
+
+	ctlplane, err := GetOpenStackNetWithLabel(
+		r,
+		namespace,
+		labelSelector,
+	)
+	if err != nil {
+		return ctlplaneIps, err
+	}
+	for val, roleReservation := range ctlplane.Status.RoleReservations {
+		roleReservations := map[string]string{}
+		for _, ipReservation := range roleReservation.Reservations {
+			roleReservations[ipReservation.Hostname] = ipReservation.IP
+		}
+		ctlplaneIps[val] = roleReservations
+	}
+	return ctlplaneIps, err
+}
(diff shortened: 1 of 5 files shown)


pkg/openstacknet/funcs.go
```diff
@@ -85,3 +85,23 @@ func GetOpenStackNetWithLabel(r common.ReconcilerCommon, namespace string, label
 	}
 	return &osNetList.Items[0], nil
 }
+
+// GetOpenStackNetworkHostnameIPs - Return a map of hostname/IPs for the specified network selector
+func GetOpenStackNetworkHostnameIPs(r common.ReconcilerCommon, namespace string, labelSelector map[string]string) (map[string]string, error) {
+	ctlplaneIps := map[string]string{}
```
stuggi, line 91: would be good if this returns a map of a struct with role -> hostname, ctlplaneip . with this we can generate in the inventory also the groups that it is easier to register the different groups to different channels, like ceph nodes

templates/openstackclient/bin/ctlplane-inventory.yaml
```diff
@@ -0,0 +1,7 @@
+overcloud:
```
stuggi, line 1: as said above would also be good to have the different host groups in the inventory to be able to subscribe them to different channels.

dprince:
@stuggi I pondered this approach a bit more yesterday and I now believe it is the wrong way forward. Instead of "manually" executing RHSM installation it would work much more nicely to just use the automated t-h-t templates for RHSM registration: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.2/html-single/advanced_overcloud_customization/index#deploying-the-overcloud-with-the-rhsm-composable-service-portal Thoughs? I will rework a new patch for dev-tools which can leverage Ollie's latest heat environment patch to implement this approach.

stuggi:
> @stuggi I pondered this approach a bit more yesterday and I now believe it is the wrong way forward. Instead of "manually" executing RHSM installation it would work much more nicely to just use the automated t-h-t templates for RHSM registration: https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.2/html-single/advanced_overcloud_customization/index#deploying-the-overcloud-with-the-rhsm-composable-service-portal Thoughs? I will rework a new patch for dev-tools which can leverage Ollie's latest heat environment patch to implement this approach.

@dprince this won't work without a change to tht, or we switch to the overcloud full image for the vmset/bmset nodes. Before the rhsm service runs, the common roles get triggered, which runs the tripleo-bootstrap role [1]. This includes installing the required ooo packages, which then won't install the required packages and later steps fail, e.g. missing network-scripts package. This will work if we'd reuse the overcloud full image which already has all the packages available.

[1] https://github.com/openstack/tripleo-heat-templates/blob/stable/train/common/deploy-steps.j2#L477

PR #485, opened 2022-01-07
Backup/restore KUTTL tests

abays:
Also fixes a few minor bugs related to backup/restore

### DIFF
--- a/controllers/openstackclient_controller.go
+++ b/controllers/openstackclient_controller.go
@@ -96,7 +96,7 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	//
 	// initialize condition
 	//
-	cond := &ospdirectorv1beta1.Condition{}
+	cond := instance.Status.Conditions.InitCondition()
 
 	if instance.Status.OpenStackClientNetStatus == nil {
 		instance.Status.OpenStackClientNetStatus = map[string]ospdirectorv1beta1.HostStatus{}
@@ -139,7 +139,7 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}(cond)
 
 	// If we determine that a backup is overriding this reconcile, requeue after a longer delay
-	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, true)
+	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, instance.Status.Conditions.GetCurrentCondition().Reason == ospdirectorv1beta1.OsClientCondReasonPodProvisioned)
 
 	if err != nil {
 		return ctrl.Result{}, err
--- a/controllers/openstackconfiggenerator.go
+++ b/controllers/openstackconfiggenerator.go
@@ -121,7 +121,7 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req 
 	//
 	// initialize condition
 	//
-	cond := &ospdirectorv1beta1.Condition{}
+	cond := instance.Status.Conditions.InitCondition()
 
 	//
 	// Used in comparisons below to determine whether a status update is actually needed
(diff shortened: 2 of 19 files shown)


controllers/openstackclient_controller.go
```diff
@@ -96,7 +96,10 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	//
 	// initialize condition
 	//
-	cond := &ospdirectorv1beta1.Condition{}
+	cond := instance.Status.Conditions.GetCurrentCondition()
+	if cond == nil {
+		cond = &ospdirectorv1beta1.Condition{}
+	}
```
stuggi, line 102: how about to create a instance.Status.Conditions.InitCondition() which does these tasks? I mean GetCurrentCondition() and if nil returns &ospdirectorv1beta1.Condition{}

controllers/openstackclient_controller.go
```diff
@@ -139,7 +142,7 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}(cond)
 
 	// If we determine that a backup is overriding this reconcile, requeue after a longer delay
-	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, true)
+	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, instance.Status.Conditions.GetCurrentCondition() != nil && instance.Status.Conditions.GetCurrentCondition().Reason == ospdirectorv1beta1.OsClientCondReasonPodProvisioned)
```
stuggi, line 145: with the above `cond := instance.Status.Conditions.GetCurrentCondition()` shouldn't it be ok to just check cond.Reason? it should never be nil here, right?

controllers/openstackconfiggenerator.go
```diff
@@ -121,7 +121,10 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
 	//
 	// initialize condition
 	//
-	cond := &ospdirectorv1beta1.Condition{}
+	cond := instance.Status.Conditions.GetCurrentCondition()
+	if cond == nil {
+		cond = &ospdirectorv1beta1.Condition{}
+	}
```
stuggi, line 127: could then use in all those just instance.Status.Conditions.InitCondition()

controllers/openstackvmset_controller.go
```diff
@@ -223,7 +226,7 @@ func (r *OpenStackVMSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 
 	// If we determine that a backup is overriding this reconcile, requeue after a longer delay
-	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, instance.Status.ProvisioningStatus.State == ospdirectorv1beta1.VMSetCondTypeProvisioned)
+	overrideReconcile, err := ospdirectorv1beta1.OpenStackBackupOverridesReconcile(r.Client, instance.Namespace, instance.Status.ProvisioningStatus.State == ospdirectorv1beta1.VMSetCondTypeProvisioned || instance.Status.ProvisioningStatus.State == ospdirectorv1beta1.VMSetCondTypeEmpty)
```
stuggi, line 229: as we discussed a while back to get rid of the dedicated ProvisioningStatus, could use cond here which is a this point currentCondition?

controllers/openstackclient_controller.go
abays: Good idea

controllers/openstackclient_controller.go
abays: Yep, and that will be cleaner to read as well

controllers/openstackconfiggenerator.go
abays: Ack

controllers/openstackvmset_controller.go
abays: That is the proper approach.  Will fix.

controllers/openstackclient_controller.go
abays: Done

controllers/openstackclient_controller.go
abays: Done

controllers/openstackvmset_controller.go
abays: Done

stuggi (approved):
/lgtm

PR #486, opened 2022-01-10
[OSPK8-424] Fix prov server interface override

abays:
For the `OpenStackProvisionServer` controller, we were caching the first interface used (whether supplied via a CR or discovered from Metal3).  This prevented subsequent CRs with an `interface` override in the spec from properly honoring and using that chosen interface.


stuggi (approved):
/lgtm

PR #489, opened 2022-01-12
[OSPK8-421] Move OsMacAddr mgmt to osnetcfg

stuggi:
This moves the osmac reservations create to the osnetcfg controller
and updates the osmacaddr spec with new/changed reservations. For
this the osvmset and osbmset controller adds a label to its CR with
the osnetcfg CR name. This allows the osnetcfg to watch osbms and
osvmset CRs and reconciles on changes. It always recreates the full
list of reservations. Therefore no special handling is required
to cleanup reservations. If a node is gone from the osbms/osvmset
status also the mac reservation is gone.

The status of the osmacaddr CR still holds the same flattened
reservation list. This is still the source to render the ooo
parameter file.

This also changes the osbms to use the same common HostStatus type
to be able to use a common function to check the created osbms
and osvmset nodes. Right now required parameters from osbms got
merged into the existing HostStatus struct. In a follow up this
should be cleaned up (CtlplaneIP).


api/v1beta1/openstackmacaddress_types.go
```diff
@@ -88,12 +98,14 @@ const (
 	MACCondReasonCreateMACError MACReason = "CreateMACError"
 	// MACCondReasonAllMACAddressesCreated - all MAC addresses created
 	MACCondReasonAllMACAddressesCreated MACReason = "MACAddressesCreated"
+	// MACCondReasonError - General error getting the OSMACaddr object
+	MACCondReasonError VMSetProvisioningReason = "MACError"
```
abays, line 102: s/VMSetProvisioningReason/MACReason

controllers/openstackbaremetalset_controller.go
```diff
@@ -1397,11 +1407,20 @@ func (r *OpenStackBaremetalSetReconciler) createNewHostnames(
 			return newHostnames, err
 		}
 
+		if instance.Status.BaremetalHosts == nil {
+			instance.Status.BaremetalHosts = map[string]ospdirectorv1beta1.HostStatus{}
+		}
```
abays, line 1412: It looks like we're already doing this on lines 121-123.  Or maybe it is possible for the instance to arrive at the `createNewHostnames` func without hitting those lines above?

controllers/openstackvmset_controller.go
```diff
@@ -1124,6 +1123,10 @@ func (r *OpenStackVMSetReconciler) createNewHostnames(
 			return newHostnames, err
 		}
 
+		if instance.Status.VMHosts == nil {
+			instance.Status.VMHosts = map[string]ospdirectorv1beta1.HostStatus{}
+		}
```
abays, line 1128: Same question here regarding this initialization as with BaremetalSet controller

controllers/openstackbaremetalset_controller.go
stuggi: yeah, I have still seen a assignment to entry in nil map error. need to debug why, or just add them here for safety

controllers/openstackvmset_controller.go
stuggi: same

PR #493, opened 2022-01-18
[OSPK8-426] Adds custom reservations and preserve MAC reservations if configured

stuggi:
This allows to add custom ovnBridgeMacMappings per node and also
preserves reservations on nodes delete per default. Its possible to
configure to not preserve reservations when a node get deleted.

Also introduce webhook checks to validate that:
* check if the static MAC address has a valid format
* check for duplicate reservations on static reservations
* check that a MAC reservation won't change


controllers/openstacknetconfig_controller.go
```diff
@@ -940,3 +992,44 @@ func (r *OpenStackNetConfigReconciler) ensureMACReservation(
 
 	return nodeMACReservation, nil
 }
+
+//
+// allMACReservations - get all reservations from static + dynamic created
+// also check if there are any dupe in OVNBridgeMacMappings.StaticReservations
```
bogdando, line 998: this doesn't look like checking for dups? also it always returns nil as error

stuggi: right, I moved the dupe check to the webhook. let me update this here

abays (approved):
/lgtm

PR #494, opened 2022-01-19
Restructure kuttl tests to remove redundancy

stuggi:
Remove redundancy in common multiple used tests/assert, e.g.
network create/ctlplane scale/...
Those can be linked into the actual test directory. This
has the advantage if a CRD changes only files in a single
location need to be updated.


abays (approved):
Very nice!

/lgtm

PR #495, opened 2022-01-24
[OSPK8-431] Migrate osipset functionality to osnetcfg

stuggi:
This migrates the IP reservation functionality from osipset to
osmacaddr controller. The osmacaddr controller watches osctlplane
osvmset, osclient and osbmset resources and creates IP reservations
if needed. The reservations get stored on the osnet spec, like before.
The ip set information is stored on the status of the osnetcfg CR.

If a node gets deleted, the default is to persist the reservations
of a node if the full role isn't deleted (like before). With this
change this can be controlled by the osnetcfg Spec.PreserveReservations
parameter, which controlls both IP and MAC address reservations.

In a follow up static IP reservation will be added.


api/v1beta1/openstacknet_types.go
```diff
@@ -94,7 +100,11 @@ type OpenStackNetSpec struct {
 
 	// +kubebuilder:validation:Optional
 	// Reservations, IP address reservations per role
-	RoleReservations map[string]OpenStackNetRoleStatus `json:"roleReservations"`
+	RoleReservations map[string]OpenStackNetRoleReservation `json:"roleReservations"`
+
+	// +kubebuilder:validation:Optional
+	// Reservations, IP address reservations per role
+	//Reservations map[string]OpenStackNetRoleReservation `json:"reservations"`
```
dprince, line 107: comment

api/v1beta1/openstacknetconfig_types.go
```diff
@@ -189,6 +203,15 @@ const (
...
 	NetConfigError ProvisioningState = "Error"
+
+	// NetConfigCondReasonWaitingOnIPsForHost - waiting on IPs for all configured networks to be created
+	NetConfigCondReasonWaitingOnIPsForHost ConditionReason = "WaitingOnIPsForHost"
+	// NetConfigCondReasonWaitingOnHost - waiting on host to be added to osnetcfg
+	NetConfigCondReasonWaitingOnHost ConditionReason = "WaitingOnHost"
+	// NetConfigCondReasonIPReservationError - Failed to do ip reservation
+	NetConfigCondReasonIPReservationError ConditionReason = "IPReservationError"
+	// NetConfigCondReasonIPReservation - ip reservation created
+	NetConfigCondReasonIPReservation ConditionReason = "IPReservatioCreated"
```
abays, line 214: Small typo: `IPReservatioCreated` needs an `n`

controllers/openstackbaremetalset_controller.go
```diff
@@ -148,10 +145,9 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
 
 		if statusChanged() {
 			if updateErr := r.Client.Status().Update(context.Background(), instance); updateErr != nil {
+				err = common.WrapErrorForObject(
+					"Update Status", instance, updateErr)
 				if err == nil {
```
abays, line 150: Will `err` ever be `nil` here?  Since `updateErr` is not `nil` and gets passed to `WrapErrorForObject`, it will be return by that func as non-`nil` `err`.

stuggi: right that should be != nil

stuggi: or just use LogErrorForObject

controllers/openstackbaremetalset_controller.go
```diff
@@ -316,22 +314,40 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
...
+	}
+
+	//
+	// Wait for IPs created on all configured networks
+	//
+	for hostname, hostStatus := range instance.Status.BaremetalHosts {
+		err = openstacknetconfig.WaitOnIPsCreated(
+			instance,
+			cond,
+			osnetcfg,
```
abays, line 340: Couldn't `osnetcfg` be an empty `OpenStackNetConfig` if the `err` above was `IsNotFound`?  If so, would `WaitOnIPsCreated` behave as expected if it received the empty object?

stuggi: good catch, we should wait until osnetcfg is there. it does not make sense to proceed as we don't have networks

stuggi: ah wait AddOSNetConfigRefLabel [1] at the top of the controller is waiting for all osnet to be available. so the osnetcfg should be there. but lets just check for all osnetcfg get errors here.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/pkg/openstacknetconfig/funcs.go#L47

stuggi: changed

pkg/openstacknet/funcs.go
```diff
@@ -85,3 +91,173 @@ func GetOpenStackNetWithLabel(r common.ReconcilerCommon, namespace string, label
...
+			obj,
+		)
+	}
+
+	//
+	// identify if nets got added
+	//
+	for label := range networkNameLowerNamesMap {
+		//
+		// if l is not in networkNameLowerNamesMap it got removed
```
abays, line 183: Is this comment correct?

pkg/openstacknet/funcs.go
```diff
@@ -85,3 +91,173 @@ func GetOpenStackNetWithLabel(r common.ReconcilerCommon, namespace string, label
...
+				obj.GetLabels(),
+				currentLabels,
+			))
+
+		return nil
+	})
+	if err != nil {
+		cond.Message = fmt.Sprintf("Failed to update %s labels on %s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.CommonCondReasonAddOSNetLabelError)
+		cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.ConfigGeneratorCondTypeError)
```
abays, line 219: Is `ConfigGenerator` the right `TypeError` here?

pkg/openstacknet/funcs.go
stuggi: oh, this is something I stopped working on for now and do in a follow up to be able to remove subnets

pkg/openstacknet/funcs.go
stuggi: ah no, but was wondering if we should have generic Types as we have Error,Creating, Created,... on all of the objects. maybe they could just use all CommonType, like we already have CommonCondTypeError? We likely never change an Error to be named different between the CRDs?

pkg/openstacknet/funcs.go
abays: Ack

pkg/openstacknet/funcs.go
abays: That seems reasonable to me

abays (approved):
/lgtm

pkg/openstacknet/funcs.go
stuggi: updated to reflect current state

pkg/openstacknet/funcs.go
stuggi: done for this one

api/v1beta1/openstacknetconfig_types.go
```diff
@@ -104,6 +104,12 @@ type Network struct {
 	// +kubebuilder:default=true
 	// VIP create virtual ip on the network
 	VIP bool `json:"vip"`
+
+	/*
+		// +kubebuilder:validation:Optional
+		// StaticReservations, manual/static IP address reservations per node
+		StaticReservations map[string]OpenStackMACNodeReservation `json:"staticReservations"`
```
olliewalsh, line 111: should this be removed?

olliewalsh (approved):

api/v1beta1/openstacknetconfig_types.go
stuggi: we could, but want to work on the static ip reservations as one of my next things and kept it as a reminder :)

dprince (approved):

PR #497, opened 2022-01-27
[OSPK8-428 OSPK8-429] allow CPU and MEM resize on vmset

stuggi:
Allows to change CPU/RAM assigned to virtual machines when changing the virtualMachineRole in the osctlplane CR.
Also now displays the configured CPU/RAM in the summary when display all vmsets

~~~
# oc get osvmset
NAME         CORES   RAM   DESIRED   READY   STATUS        REASON
controller   6       21    1         1       Provisioned   All requested VirtualMachines have been provisioned
~~~


tests/kuttl/tests/openstackcontrolplane_cpu_ram_scale/04-assert.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: TestAssert
+timeout: 180
+commands:
+  - script: |
+      #!/usr/bin/env bash
+      sleep 10
+      CONTROLLER=$(oc get -n openstack osnet ctlplane -o json | jq .status.reservations | jq -r "to_entries | map(select(.key == \"controller-0\")) | unique[] | .value.ip")
+      CPUS=$(oc rsh -n openstack openstackclient ssh $CONTROLLER "grep processor /proc/cpuinfo | wc -l")
+      RAM=$(oc rsh -n openstack openstackclient ssh $CONTROLLER grep MemTotal /proc/meminfo | awk '{ print $2 }')
+      if [[ ( "${CPUS}" != 6 ) || ( "${RAM}" != 20340536 ) ]]; then
```
olliewalsh, line 18: I don't think the RAM check will be reliable. The MemTotal value is always less than the physical RAM. I expect it will vary depending on the libvirt/qemu version (bios, video device etc...) and the kernel version on the VM.

Maybe just check that it's 19-20GB and then check that 20-21GB when the RAM has been increased?

README.md
```diff
@@ -1543,3 +1543,24 @@ openstackbackuprestore   restore    openstackbackupsave-1641928378   Restored
...
+```bash
+oc patch -n openstack osctlplane overcloud --type='json' -p='[{"op": "add", "path": "/spec/virtualMachineRoles/controller/cores", "value": 8 }]'
+oc patch -n openstack osctlplane overcloud --type='json' -p='[{"op": "add", "path": "/spec/virtualMachineRoles/controller/memory", "value": 21 }]'
+
+oc get osvmset
+NAME         CORES   RAM   DESIRED   READY   STATUS        REASON
+controller   8       21    1         1       Provisioned   All requested VirtualMachines have been provisioned
+```
+
+* schedule a restart of the virtual machine to get the change reflected inside the virtual machine (**Important** it is required to power off/on the virtual machine)
```
abays, line 1566: Should we mention something about `virtctl restart <VM>` here in regards to this requirement?

tests/kuttl/tests/openstackcontrolplane_cpu_ram_scale/04-assert.yaml
stuggi: ack

README.md
stuggi: I had a broken file system when using virtctl restart. therefore graceful shutdown from inside the VM and virtctl start is probable better

olliewalsh: virtctl restart should be graceful, are we missing a guest agent on the VM?

stuggi: no its there, but I think its because we set terminationGracePeriodSeconds: 0 so that when there is a fencing action triggered they get immediately switched off.

tests/kuttl/tests/openstackcontrolplane_cpu_ram_scale/04-assert.yaml
stuggi: changed the test

olliewalsh: hrmm, github has not updated the PR

README.md
```diff
@@ -1543,3 +1543,24 @@ openstackbackuprestore   restore    openstackbackupsave-1641928378   Restored
...
+```bash
+oc patch -n openstack osctlplane overcloud --type='json' -p='[{"op": "add", "path": "/spec/virtualMachineRoles/controller/cores", "value": 8 }]'
+oc patch -n openstack osctlplane overcloud --type='json' -p='[{"op": "add", "path": "/spec/virtualMachineRoles/controller/memory", "value": 22 }]'
+
+oc get osvmset
+NAME         CORES   RAM   DESIRED   READY   STATUS        REASON
+controller   8       22    1         1       Provisioned   All requested VirtualMachines have been provisioned
+```
+
+* schedule a restart of the virtual machine to get the change reflected inside the virtual machine (**Important** it is required to power off/on the virtual machine). The recommended way is to do a graceful shutdown from inside the virtual machine and use `virtctl start <VM>` to power the VM back on.
```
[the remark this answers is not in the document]
stuggi: a graceful shutdown should not trigger any fencing action

dprince (approved):

PR #498, opened 2022-01-28
[OSPK8-432] Prevent NNCP updates while configuration is in progress

stuggi:
Randomly the following error was seen:

{"level":"info","ts":1643296206.0834217,"logger":"enactmentstatus","msg":"status: {DesiredState:interfaces:\n- bridge:\n    options:\n      stp:\n        enabled: false\n    port:\n    - name: enp6s0\n  description: Linux bridge with enp6s0 as a port\n  mtu: 1500\n  name: br-ex\n  state: up\n  type: linux-bridge\n
...
ERROR: State editing already in progress.\nCommit, roll back or wait before retrying.\n' LastHeartbeatTime:2022-01-27 15:10:06.083365961 +0000 UTC m=+5.790948854 LastTransitionTime:2022-01-27 15:10:06.083365961 +0000 UTC m=+5.790948854}

We suspect that it is could be a result of applying an nncp update
while nmstate is sill applying the config. This patch prevents
the osnetattach controller to update the NNCP CR when the NNCP CR
is not in SuccessfullyConfigured ConditionReason.

### DIFF
--- a/api/v1beta1/common_types.go
+++ b/api/v1beta1/common_types.go
@@ -98,6 +98,8 @@ const (
 	CommonCondReasonNewHostnameError ConditionReason = "NewHostnameError"
 	// CommonCondReasonCRStatusUpdateError - error updating CR status
 	CommonCondReasonCRStatusUpdateError ConditionReason = "CRStatusUpdateError"
+	// CommonCondReasonNNCPError - NNCP error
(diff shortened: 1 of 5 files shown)


abays (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #499, opened 2022-01-31
[OSPK8-434] Harvest Ctlplane exports from heat

dprince:
A step towards DCN support

Updates the ConfigGeneratorController with logic to harvest
controlPlane Heat parameter exports from the Heat stack.
These are subsequently saved alongside of the ConfigVersion
and will be combined with the overcloud.json data (which
gets generated after deployment) to create the final
set of Ctlplane heat parameter exports.

Also fixes an issue with the ConfigGenerator so that we can create
multiple: previously the ConfigMap name was hard coded preventing
multiple ConfigGenerators from being created.

### DIFF
--- a/controllers/openstackconfiggenerator_controller.go
+++ b/controllers/openstackconfiggenerator_controller.go
@@ -367,6 +367,7 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req 
 	// Define a new Job object
 	job := openstackconfiggenerator.ConfigJob(instance, configMapHash, OSPVersion)
 
+	var exports string
 	if instance.Status.ConfigHash != configMapHash {
 		op, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, heat, func() error {
 			err := controllerutil.SetControllerReference(instance, heat, r.Scheme)
@@ -507,6 +508,17 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req 
 
 			return ctrl.Result{RequeueAfter: time.Second * 10}, nil
 		}
+		// obtain the cltplaneExports from Heat
+		exports, err = openstackconfiggenerator.CtlplaneExports("heat-"+instance.Name, r.Log)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			cond.Message = err.Error()
+			cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.ConfigGeneratorCondReasonEphemeralHeatDelete)
+			cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.ConfigGeneratorCondTypeError)
+			err = common.WrapErrorForObject(cond.Message, instance, err)
+
+			return ctrl.Result{}, err
+		}
+
 	}
 
 	r.setConfigHash(instance, configMapHash)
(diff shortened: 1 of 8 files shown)


controllers/openstackconfiggenerator_controller.go
```diff
@@ -521,11 +521,22 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
+	// obtain the cltplaneExports from Heat
+	exports, err := openstackconfiggenerator.CtlplaneExports("heat-"+instance.Name, r.Log)
```
dprince, line 525: FIXME: Need to run this only conditionally *if* the stack is for Ctlplanes (has Controllers)

dprince:
Hacked on this over the weekend. Its close but currently fails with:

Reconciler error        {"reconciler group": "osp-director.openstack.org", "reconciler kind": "OpenStackConfigGenerator", "name": "default", "namespace": "openstack", "error": "Get \"http://heat-default/v1/admin/stacks/overcloud\": dial tcp 172.30.123.61:80: connect: no route to host *v1beta1.OpenStackConfigGenerator openstack/default: Get \"http://heat-default/v1/admin/stacks/overcloud\": dial tcp 172.30.123.61:80: connect: no route to host"}

The URL is accessible from other pods aside from the manager. Or I could expose it via a route I guess but that seems like overkill

pkg/openstackconfiggenerator/export.go
```diff
@@ -0,0 +1,77 @@
...
+		"EndpointMap":  "EndpointMapOverride",
+		"HostsEntry":   "ExtraHostFileEntries",
+		"GlobalConfig": "GlobalConfigExtraMapData",
+	}
+}
+
+// CtlplaneExports -
+func CtlplaneExports(heatServiceName string, log logr.Logger) (string, error) {
+
+	provider, err := openstack.NewClient("http://" + heatServiceName + "/")
```
dprince, line 38: Needs to be port 8004

stuggi (approved):
/lgtm

controllers/openstackconfiggenerator_controller.go
```diff
@@ -525,7 +537,7 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
 	err = r.Client.Delete(context.TODO(), heat)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		cond.Message = err.Error()
-		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.ConfigGeneratorCondReasonEphemeralHeatDelete)
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.ConfigGeneratorCondReasonExportFailed)
```
abays, line 540: Should this be swapped with line 515, since this is a delete failure and not an export failure?

olliewalsh (approved):
/lgtm

pkg/openstackconfiggenerator/export.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks"
+)
+
+func exportNames() map[string]string {
+	return map[string]string{
+		"EndpointMap":  "EndpointMapOverride",
+		"HostsEntry":   "ExtraHostFileEntries",
+		"GlobalConfig": "GlobalConfigExtraMapData",
+	}
+}
```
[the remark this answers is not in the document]
dprince: You mean like a private variable. Sure. It cant be a constant however.

controllers/openstackconfiggenerator_controller.go
dprince: Yes. Not really sure how I made that happen. Will fix in the next commit.

pkg/openstackconfiggenerator/export.go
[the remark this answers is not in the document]
dprince: I guess (initially) I was thinking of a public function. But since I'm only using it locally a private variable works just fine.

PR #500, opened 2022-01-31
[OSPK8-433] Allow to set StorageAccessMode and StorageVolumeMode for VMset

stuggi:
This patch allows to set the StorageAccessMode and/or StorageVolumeMode
for the VirtualMachineRole disk. StorageAccessMode defaults to
ReadWriteMany as this is required to be able to live migrate
an instance, but could be set to ReadWriteOnce if wanted.
StorageVolumeMode defaults to Filesystem, but when OpenShift
Virtualization with OpenShift Container Platform Container Storage
is used, Block should be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,7 +482,14 @@ Create a base RHEL data volume prior to deploying OpenStack.  This will be used 
           memory: 12
           diskSize: 50
           baseImageVolumeName: openstack-base-img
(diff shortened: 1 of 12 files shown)


abays (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #505, opened 2022-02-01
Ensure ansible playbooks symlink is correct

olliewalsh:
If the symlink target already exists then a symlink is created within it
instead of updating the symlink source. This could result in the wrong
playbooks being used.


stuggi (approved):
/lgtm

PR #506, opened 2022-02-01
Add Role param to ConfigGenerator

dprince:
This can be used to explicitly include on the roles you
want to generate. It disables resource checks and includes only
network info for the roles specified.

NOTE: Using this feature likely requires customizations
to roles_data.yaml in the tripleo-heat-templates as well.


pkg/openstackconfiggenerator/configmap.go
```diff
@@ -347,6 +347,23 @@ func createNetworksMap(
 	return networksMap, networkMappingList, nil
 }
 
+// IsRoleIncluded - checks if the role exists in the ConfigGenerator Roles set
+func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator) bool {
+
+	if len(instance.Spec.Roles) == 0 {
+		return true
+	} else if roleName == "ControlPlane" {
```
stuggi, line 355: should use openstackcontrolplane.Role const [1] to not have to hard code the string.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/9a77ee2c7abe10c11f20622ccdbc3a858952226f/pkg/controlplane/const.go#L30

pkg/openstackconfiggenerator/configmap.go
```diff
@@ -347,6 +347,23 @@ func createNetworksMap(
 	return networksMap, networkMappingList, nil
 }
 
+// IsRoleIncluded - checks if the role exists in the ConfigGenerator Roles set
+func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator) bool {
+
+	if len(instance.Spec.Roles) == 0 {
```
stuggi, line 353: could combine in one if for check on len and ctlplane role?

dprince: done

pkg/openstackconfiggenerator/configmap.go
dprince: done

PR #507, opened 2022-02-01
[OSPK8-437] Move provisioningAgent into common agent

dprince:
We now have a common "agent" container image which can
be used for multiple commands. This patch moves
the provisioning-server-ip-discovery container
command into the common agent image. It also
renames the parameter to reference the more
generic agentImageUrl.

The goal is to decrease the number of container image deliverables
we need to maintain.


abays (approved):
/lgtm

PR #508, opened 2022-02-02
Drop GitSecret params from OpenStackClient CRD

dprince:
Now that we have the OpenStackDeploy CR workflow we no
longer need to specify GitSecret on the OpenStackClient
and OpenStackControlPlane CRs.

OSPK8-438 #close Closing issue.


dprince:
> Just tested and this currently fails. Looking into it...

Just pushed an update. Should resolve the issue.

stuggi (approved):
/lgtm

PR #509, opened 2022-02-03
[OSPK8-415] Deferred status pattern for prov server

abays:
Switches provision server controller to use the new deferred status approach

### DIFF
--- a/api/v1beta1/openstackprovisionserver_types.go
+++ b/api/v1beta1/openstackprovisionserver_types.go
@@ -69,17 +69,29 @@ const (
 	ProvisionServerCondTypeError ProvisioningState = "Error"
 
 	//
-	// condition reasones
+	// condition reasons
 	//
 
 	// OpenStackProvisionServerCondReasonListError - osprovserver list objects error
 	OpenStackProvisionServerCondReasonListError ConditionReason = "OpenStackProvisionServerListError"
+	// OpenStackProvisionServerCondReasonGetError - osprovserver list objects error
+	OpenStackProvisionServerCondReasonGetError ConditionReason = "OpenStackProvisionServerCondReasonGetError"
 	// OpenStackProvisionServerCondReasonNotFound - osprovserver object not found
 	OpenStackProvisionServerCondReasonNotFound ConditionReason = "OpenStackProvisionServerNotFound"
-	// OpenStackProvisionServerCondReasonNotReady - osprovserver not yet ready
-	OpenStackProvisionServerCondReasonNotReady ConditionReason = "OpenStackProvisionServerNotReady"
-	// OpenStackProvisionServerCondReasonReady - osprovserver ready
-	OpenStackProvisionServerCondReasonReady ConditionReason = "OpenStackProvisionServerReady"
+	// OpenStackProvisionServerCondReasonInterfaceAcquireError - osprovserver hit an error while finding provisioning interface name
+	OpenStackProvisionServerCondReasonInterfaceAcquireError ConditionReason = "OpenStackProvisionServerCondReasonInterfaceAcquireError"
+	// OpenStackProvisionServerCondReasonInterfaceNotFound - osprovserver unable to find provisioning interface name
+	OpenStackProvisionServerCondReasonInterfaceNotFound ConditionReason = "OpenStackProvisionServerCondReasonInterfaceNotFound"
+	// OpenStackProvisionServerCondReasonDeploymentError - osprovserver associated deployment failed to create/update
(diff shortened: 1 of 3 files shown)


api/v1beta1/openstackprovisionserver_types.go
```diff
@@ -69,13 +69,29 @@ const (
...
+	// OpenStackProvisionServerCondReasonDeploymentError - osprovserver associated deployment failed to create/update
+	OpenStackProvisionServerCondReasonDeploymentError ConditionReason = "OpenStackProvisionServerCondReasonDeploymentError"
+	// OpenStackProvisionServerCondReasonDeploymentCreated - osprovserver associated deployment has been created/update
+	OpenStackProvisionServerCondReasonDeploymentCreated ConditionReason = "OpenStackProvisionServerCondReasonDeploymentCreated"
+	// OpenStackProvisionServerCondReasonProvisioning - osprovserver associated pod is provisioning
+	OpenStackProvisionServerCondReasonProvisioning ConditionReason = "OpenStackProvisionServerCondReasonProvisioning"
+	// OpenStackProvisionServerCondReasonLocalImageURLParseError - osprovserver was unable to parse its received local image URL
+	OpenStackProvisionServerCondReasonLocalImageURLParseError ConditionReason = "OpenStackProvisionServerCondReasonLocalImageURLParseError"
+	// OpenStackProvisionServerCondReasonProvisioned - osprovserver associated pod is provisioned
+	OpenStackProvisionServerCondReasonProvisioned ConditionReason = "OpenStackProvisionServerCondReasonProvisioned"
```
abays, line 94: @stuggi I wonder if it might be better to combine these `Provision*` reasons with your existing `*Ready` and `Created` ones, or replace one with the other, etc

controllers/openstackprovisionserver_controller.go
```diff
@@ -204,49 +216,71 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
...
-			actualProvisioningState.Reason = err.Error()
-			_ = r.setProvisioningStatus(instance, actualProvisioningState)
+		if err != nil {
+			if k8s_errors.IsNotFound(err) {
+				cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.ProvisionServerCondTypeWaiting)
+				cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.OpenStackProvisionServerCondReasonListError)
+				cond.Message = "List of pods not found"
+				err = common.WrapErrorForObject(cond.Message, instance, err)
+
+				return ctrl.Result{}, err
```
bogdando, line 246: shouldn't it RequeueAfter?..

controllers/openstackprovisionserver_controller.go
```diff
@@ -204,49 +216,71 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
...
-			// There should only be one pod.  If there is more than one, we have other problems...
-			for _, pod := range podList.Items {
-				if pod.Status.Phase == corev1.PodRunning {
-					actualProvisioningState.State = ospdirectorv1beta1.ProvisionServerCondTypeProvisioned
-					actualProvisioningState.Reason = fmt.Sprintf("ProvisionServer %s has been provisioned", instance.Name)
-					break
-				}
+		}
+
+		// There should only be one pod.  If there is more than one, we have other problems...
```
bogdando, line 257: maybe log a warning if multiple pods found?

controllers/openstackprovisionserver_controller.go
abays: Either that or remove this specific error check entirely.  In hindsight, I'm not even sure the `Pods.List()` func can even return an `IsNotFound` error.  It should always return a list, it just might be empty.  So perhaps the proper thing to do is to check the list length after checking for any error in general.

controllers/openstackprovisionserver_controller.go
abays: Good idea, added

controllers/openstackprovisionserver_controller.go
abays: I tested the `List` func over the weekend and found that it does not return a `NotFound` error (in contradistinction to the `Get` func).  So I removed the check for `NotFound` and moved that consideration to a block I added to analyze the returned-list's length.

stuggi (commented):
other then what you suggested to maybe combine the reasons it lgtm

abays:
> other then what you suggested to maybe combine the reasons it lgtm

I merged the `NotReady` and `Ready` reasons with `Provisioning` and `Provisioned`.  I think that serves to address the combination concern.

stuggi (approved):
/lgtm

PR #510, opened 2022-02-03
[OSPK8-327] Allow static ip reservation

stuggi:
This allows to do static IP reservation for nodes by adding a
node entry to spec.reservations, like:

```
  reservations:
    controller-0:
      ipReservations:
        ctlplane: 192.168.25.40
      macReservations: {}
    controller-1:
      ipReservations:
        ctlplane: 192.168.25.41
      macReservations: {}
    controller-2:
      ipReservations:
        ctlplane: 192.168.25.42
      macReservations: {}
    controlplane:
      ipReservations:
        ctlplane: 192.168.25.10
      macReservations: {}
    openstackclient-0:
      ipReservations:
        ctlplane: 192.168.25.251
      macReservations: {}
```

Where node name is RoleName lower + index. For the VIPs of the control
plane use controlplane and openstackclient-0 for the openstackclient
pod.

Via webhook it gets verified if
* there are duplicates in the reservatios list to prevent that static
IPs get used for multiple hosts. In a follow up it should also get
checked if there are duplicates on update from static IPs and dynamic
created.
* if the IP address from static reservations have a valid IPv4/6
format.

Adds openstacknetconfig_ip_reservation kuttl tests.

Also fixes a race in IP address generation due to different sources used.


api/v1beta1/openstacknetconfig_webhook.go
```diff
@@ -247,14 +275,79 @@ func (r *OpenStackNetConfig) validateStaticMacReservations(old runtime.Object) e
...
+			// check if IP matches osnet spec
+			//
+			for _, osNet := range r.Spec.Networks {
+				for _, subnet := range osNet.Subnets {
+					if subnet.Name == netName {
+						var ipnet *net.IPNet
+						if subnet.IPv4.Cidr != "" {
+							_, ipnet, _ = net.ParseCIDR(subnet.IPv4.Cidr)
+						} else {
+							_, ipnet, _ = net.ParseCIDR(subnet.IPv4.Cidr)
```
abays, line 318: Should this be `IPv6`?

controllers/openstackbaremetalset_controller.go
```diff
@@ -343,7 +344,7 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
 			&hostStatus,
 		)
 		if err != nil {
-			return ctrl.Result{RequeueAfter: 10 * time.Second}, err
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
abays, line 347: Is the `err` removed here because including it prevents the requeue after 10 seconds?

pkg/openstacknetconfig/funcs.go
```diff
@@ -118,18 +119,27 @@ func WaitOnIPsCreated(
...
+			fmt.Sprintf("%s %s waiting on node %s to be added to %s config %s",
+				obj.GetObjectKind().GroupVersionKind().Kind,
+				obj.GetName(),
+				hostname,
+				osnetcfg.Kind,
+				osnetcfg.Name,
+			),
+			obj,
+		)
+		cond.Message = fmt.Sprintf("%s %s waiting on IPs to be created for all nodes and networks",
```
abays, line 138: Ah, I see what you mean about keeping the `Message` more uniform across reconciles.  Definitely can envision the old form causing some thrashing.

api/v1beta1/openstacknetconfig_webhook.go
```diff
@@ -247,14 +275,79 @@ func (r *OpenStackNetConfig) validateStaticMacReservations(old runtime.Object) e
...
+		reservations[node] = OpenStackMACNodeReservation{
+			Reservations: res.MACReservations,
+		}
 	}
+
+	return nil
+}
+
+// validateStaticIPReservations - validate static IP address reservations
+func (r *OpenStackNetConfig) validateStaticIPReservations(old runtime.Object) error {
```
bogdando, line 294: the ``old`` arg looks not used in this func?

api/v1beta1/openstacknetconfig_webhook.go
stuggi: right, good catch, will fix it and add an ipv6 kuttl tests which should have shown this :)

api/v1beta1/openstacknetconfig_types.go
```diff
@@ -151,17 +151,28 @@ type OpenStackNetConfigSpec struct {
 	// but the role itself is not deleted. The reservations of all nodes in the role get deleted when the full node
 	// role is being deleted. (default: true)
 	PreserveReservations *bool `json:"preserveReservations"`
+
+	// +kubebuilder:validation:Optional
+	// Reservations, manual/static MAC/IP address reservations per node
+	Reservations map[string]OpenStackNetStaticNodeReservations `json:"reservations"`
```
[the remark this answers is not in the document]
stuggi: marker kubebuilder:validation:Optional makes sure that the parameter is as it says optional, omitempty is not required. On the other hand if you use marker kubebuilder:validation:Required and in addition have omitempty specified, omitempty has precedence over the kubebuilder marker.

api/v1beta1/openstacknetconfig_webhook.go
stuggi: I added it as I want to extend the func to validate on update that there are no dupes on static and dynamic created ips. will enhance that now

controllers/openstackbaremetalset_controller.go
stuggi: I just removed it to make the log look cleaner and log an info message instead of an error message in that case, if you all agree :)

abays: Sounds good to me!

abays:
This seems ready to me.  Unless anyone else wants to review, let's merge it.

abays (approved):
/lgtm

PR #511, opened 2022-02-04
OSP minor update support

olliewalsh:
Allows playbook path, limit, tags, and skiptags to be set for a
deployment to support the OSP minor update workflow.
Adds a deploy "mode" alias for the deploy/update/externalUpdate playbooks.
Rework the image defaults so that the openstackclient image default applies to the openstackclient CR vs the control-plane CR - can then just remove the openstackclient CR after bumping the default images to upgrade it.


dprince (commented):
Seems like a reasonable approach to me. Still a Draft I know but should we update the config/sample for OSDeploy so that it contains these references (and perhaps a comment) about when you might use these options. Also, perhaps make a quick note in the readme.

api/v1beta1/openstackcontrolplane_webhook.go
```diff
@@ -121,59 +116,10 @@ func (r *OpenStackControlPlane) ValidateDelete() error {
...
-			openstackephemeralheatlog.Info("default", "name", r.Name)
-			ospVersion = OSPVersion(openstackControlPlaneDefaults.OpenStackRelease)
-		}
-
-		r.Spec.OpenStackRelease = string(ospVersion)
-	}
-}
-
-// GetOSPVersionFromImageURL -
-func GetOSPVersionFromImageURL(imageURL string) (OSPVersion, error) {
```
olliewalsh, line 148: @stuggi does this still make sense now that we are using digests? Guess it might still be useful for train/wallaby but we can just set it explicitly in those case

api/v1beta1/openstackdeploy_types.go
```diff
@@ -54,16 +54,43 @@ const (
 	DeployCondReasonConfigCreate ConditionReason = "ConfigCreate"
 )
 
+// OpenStackAnsibleSettingsSpec defines advanced ansible-playbook settings
+type OpenStackAnsibleSettingsSpec struct {
+	// +kubebuilder:validation:Optional
+	// Playbook to run from config-download
```
dprince, line 60: Should we also specify that it runs the deployment playbook be default if it is unset?

dprince: Also, should we validate with an Enum for all the valid options you have in the case statement below?

api/v1beta1/openstackcontrolplane_webhook.go
stuggi: we still use the version tagged URL to set the default images [1] which then points to the latest one using digest, or did I get it wrong?

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/config/default/manager_default_images.yaml#L15

olliewalsh: We should be replacing the tags with digests when building the CSV e.g https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/scripts/build_and_push_images.sh#L66

Not sure where this is done for the release CSV, @dprince is it scripts/stage_downstream_rebase.sh?

api/v1beta1/openstackdeploy_types.go
olliewalsh: > Should we also specify that it runs the deployment playbook be default if it is unset?

Sure, can explain how this works with the mode param

> Also, should we validate with an Enum for all the valid options you have in the case statement below?

I figured it would be better not to restrict the advanced ("I know what I'm doing") playbook option as I've not covered some of the other cases yet:
```
$ ls tripleo-ansible/*playbook*
tripleo-ansible/common_deploy_steps_playbooks.yaml   tripleo-ansible/external_upgrade_steps_playbook.yaml     tripleo-ansible/scale_playbook.yaml
tripleo-ansible/deploy_steps_playbook.yaml           tripleo-ansible/post_upgrade_steps_playbook.yaml         tripleo-ansible/update_steps_playbook.yaml
tripleo-ansible/external_update_steps_playbook.yaml  tripleo-ansible/pre_upgrade_rolling_steps_playbook.yaml  tripleo-ansible/upgrade_steps_playbook.yaml
```

api/v1beta1/openstackcontrolplane_webhook.go
olliewalsh: ah it's the github action - https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/.github/create_bundle.sh

dprince (approved):

PR #512, opened 2022-02-08
[OSPK8-414] Use deferred status pattern for OSEphHeat

abays:
Also fixes some condition reason/type stuff for OSBaremetalSets


api/v1beta1/common_types.go
```diff
@@ -254,7 +254,7 @@ func (conditions ConditionList) InitCondition() *Condition {
 		}
 	}
 
-	return cond
+	return cond.DeepCopy()
```
stuggi, line 257: not sure I understand why is this change is needed?

stuggi (commented):
just one comment as I don't get it right now :) -  other then that looks good to me

api/v1beta1/common_types.go
stuggi: thanks, yes that makes sense after having a closer look

stuggi (approved):
/lgtm

PR #513, opened 2022-02-15
Fix osp17 nic template and remove workaround

stuggi:
also removes dupe fencing template.

Depends-On: https://github.com/openstack-k8s-operators/osp-director-dev-tools/pull/292


abays (approved):
/lgtm

PR #517, opened 2022-02-18
Fix adding osnetcfg and osnet labels on CR

stuggi:
Right now AddOSNetConfigRefLabel and AddOSNetNameLowerLabels
updated the object when added the required labels to the metadata.
Since the generic way of the functions only updated the object
labels at that point of the reconcilation other object information,
like status got lost.
This changes the two functions to merge the required labels into
the object labels and returns them so that they can be set on the
CR object instead of writing the object in the specific label
functions.

### DIFF
--- a/controllers/openstackbaremetalset_controller.go
+++ b/controllers/openstackbaremetalset_controller.go
@@ -232,11 +232,17 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
 	}
 
 	var ctrlResult reconcile.Result
+	currentLabels := instance.DeepCopy().Labels
 	//
 	// add osnetcfg CR label reference which is used in the in the osnetcfg
 	// controller to watch this resource and reconcile
 	//
-	ctrlResult, err = openstacknetconfig.AddOSNetConfigRefLabel(r, instance, cond, instance.Spec.Networks[0])
+	instance.Labels, ctrlResult, err = openstacknetconfig.AddOSNetConfigRefLabel(
+		r,
+		instance,
+		cond,
+		instance.Spec.Networks[0],
+	)
 	if (err != nil) || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
 	}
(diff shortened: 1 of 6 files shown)


controllers/openstackbaremetalset_controller.go
```diff
@@ -1429,10 +1434,6 @@ func (r *OpenStackBaremetalSetReconciler) createNewHostnames(
 			return newHostnames, err
 		}
 
-		if instance.Status.BaremetalHosts == nil {
-			instance.Status.BaremetalHosts = map[string]ospdirectorv1beta1.HostStatus{}
-		}
-
```
stuggi, line 1435: @abays now I understood why this was required :)

controllers/openstackbaremetalset_controller.go
```diff
@@ -232,23 +232,48 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
...
+	)
 	if (err != nil) || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
 	}
 
 	//
 	// add labels of all networks used by this CR
 	//
-	err = openstacknet.AddOSNetNameLowerLabels(r, instance, cond, instance.Spec.Networks)
+	instance.Labels = openstacknet.AddOSNetNameLowerLabels(r, instance, cond, instance.Spec.Networks)
```
bogdando, line 253: did you miss an assignment to err?

controllers/openstackclient_controller.go
```diff
@@ -152,19 +152,44 @@ func (r *OpenStackClientReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	)
 	if (err != nil) || (ctrlResult != ctrl.Result{}) {
 		return ctrlResult, err
 	}
 
 	//
 	// add labels of all networks used by this CR
 	//
-	err = openstacknet.AddOSNetNameLowerLabels(r, instance, cond, instance.Spec.Networks)
+	instance.Labels = openstacknet.AddOSNetNameLowerLabels(r, instance, cond, instance.Spec.Networks)
```
bogdando, line 169: ditto

controllers/openstackbaremetalset_controller.go
stuggi: don't think so, the changed function no longer returns an error

pkg/openstacknet/funcs.go
```diff
@@ -124,11 +122,11 @@ func AddOSNetNameLowerLabels(
 	obj client.Object,
 	cond *ospdirectorv1beta1.Condition,
 	networkNameLowerNames []string,
-) error {
+) map[string]string {
```
bogdando, line 125: I think it should return an error as well, as you compare to ``err != nil`` above after the calls being made?

controllers/openstackbaremetalset_controller.go
stuggi: ah got it, can remove the error check after the function call. thanks for the pointer

controllers/openstackbaremetalset_controller.go
abays: Ah, was it a necessary to try to avoid the problem you are fixing in this PR?

stuggi: I am not sure I get the question correct, but if we do not fix it you basically can not trust parameters are set correct at places in the reconcile loop and have to validate them somehow. so I think the correct way is to fix it like in this PR

abays (approved):
/lgtm

PR #518, opened 2022-02-21
Fix duplicate ip assignmnent

stuggi:
Already assigned IPs right now are only considered as reserved when
they got added to the osnet status. If IPs for multiple roles get
created in the same run, new reservations are only added to the
osnet spec and not yet synced to the osnet status. Because of that
it might happen that IPs assigned/created for different get assigned
to nodes from another role.

This fix make sure that also new IP reservations added to the osnet
spec and not yet synced to the osnet status are considered as
reserved and therefore are not a candidate for new ip assignments.

### DIFF
--- a/pkg/openstacknet/funcs.go
+++ b/pkg/openstacknet/funcs.go
@@ -254,6 +254,31 @@ func GetAllIPReservations(
 	}
 
 	//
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #519, opened 2022-02-24
Harvest overcloud.json after deployment

dprince:
This patch harvests overcloud.json (generated by TripleO Ansible)
after a successful deployment finishes. The results are merged
into other "exported" data from the Heat stack and then written
to the 'tripleo-ctlplane-exports' ConfigMap.

Then intent is to optionally use this ConfigMap alongside
a new ConfigGenerator to implement a multi-stack workflows
like DCN.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1545,6 +1545,25 @@ openstackbackuprestore   restore    openstackbackupsave-1641928378   Restored   
 
 At this point, all resources contained with the chosen `OpenStackBackup` should be restored and fully provisioned.
 
+Exported stack data config map
+--------------------------------------------------------
+
+The OSP Director Operator automatically creates a ConfigMap after each OSDeploy resource finishes executing. This ConfigMap is named after the OSDeploy resource name and prefixed with tripleo-exports-. For example tripleo-exports-default would be the name of the ConfigMap for the 'default' OSDeploy resource. Each ConfigMap contains 2 YAML files:
+
+| Filename    | Description | TripleO Command Equivalent |
+| ----------- | ----------- | -------------------------- |
+| ctlplane-export.yaml | Used with multiple stacks for DCN | overcloud export |
+| ctlplane-export-filtered.yaml | Used for multiple stacks with Cell "Controller" stacks | overcloud cell export |
+
+Use the command below to extract the YAML files from the ConfigMap. Once extracted the YAML files can be
+added into custom Heat parameters on OSConfigGenerator resources.
+
+```bash
+oc extract cm/tripleo-exports-default
+```
+
+NOTE: The OSP Director Operator does not yet generate exports for Ceph stacks.
+
 # Day2 Operations
 
 ## Change resources on virtual machines
--- a/api/v1beta1/openstackconfigversion_types.go
+++ b/api/v1beta1/openstackconfigversion_types.go
@@ -36,7 +36,7 @@ type OpenStackConfigVersionStatus struct {
 //+kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=osconfigversion;osconfigversions
 // +operator-sdk:csv:customresourcedefinitions:displayName="OpenStack Config Version"
-// +kubebuilder:printcolumn:name="Generator",type="string",JSONPath=".spec.ConfigGeneratorName",description="Config Generator Name"
+// +kubebuilder:printcolumn:name="Generator",type="string",JSONPath=".spec.configGeneratorName",description="Config Generator Name"
 
 // OpenStackConfigVersion represents a set of executable Ansible playbooks
 type OpenStackConfigVersion struct {
--- a/containers/agent/deploy.go
+++ b/containers/agent/deploy.go
@@ -2,23 +2,30 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
 	"os"
 	"os/signal"
+	"reflect"
 	"strings"
 	"syscall"
 
 	"github.com/golang/glog"
+	"github.com/openstack-k8s-operators/osp-director-operator/pkg/openstackdeploy"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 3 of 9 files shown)


controllers/openstackconfiggenerator_controller.go
```diff
@@ -683,10 +686,32 @@ func (r *OpenStackConfigGeneratorReconciler) verifyNodeResourceStatus(instance *
...
 		}
 	}
+	hasControllersForRoles := r.isControllerGenerator(vmset, bmset)
 
 	return msg, true, nil
 }
 
+//
+// Is this ConfigGenerator creating resources for a Controller
+//
```
dprince, line 696: This part is still in progress and perhaps I should break it out into a separate patch. The intent is to use this function to only execute the overcloud.json harvesting if the stack *is* the ControlPlane (i.e. has Controller roles in it)

containers/agent/deploy.go
```diff
@@ -41,13 +49,29 @@ var (
...
+		Resource: "openstackconfigversions",
+	}
+
+	filteredAllNodesConfig = []string{
+		"oslo_messaging_notify_short_bootstrap_node_name",
+		"oslo_messaging_notify_node_names",
+		"oslo_messaging_rpc_node_names",
+		"memcached_node_ips",
+		"ovn_dbs_vip",
+		"redis_vip",
```
stuggi, line 65: wondering if this should be a parameter of the osdeploy (defaulting to this list) and then passed into the agent, or the agent get it from checking the osdeploy CR? with that this list could be easy customized in case filter requirements change.

Also filter is only done for a cell export [1] from the central controller deployment stack. In case of splitting the cell controllers and cell computes in different stacks, you use the same export like DCN, where there is no filtering [2], also not for DCN export [3]. So like we discussed in one of our previous meetings, we probably should just do both exports, one which filters and the one without?

[1] https://github.com/openstack/python-tripleoclient/blob/stable/train/tripleoclient/v1/overcloud_cell.py#L89
[2] https://github.com/openstack/python-tripleoclient/blob/stable/train/tripleoclient/v1/overcloud_cell.py#L92
[3] https://github.com/openstack/python-tripleoclient/blob/stable/train/tripleoclient/v1/overcloud_export.py#L93

containers/agent/deploy.go
```diff
@@ -93,6 +190,97 @@ func ExecPodCommand(kclient kubernetes.Clientset, pod corev1.Pod, containerName
...
+		context.TODO(),
+		deployOpts.pod,
+		metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	// obtain the allNodesData from the overcloud.json from the deployment pod
+	// this is written by TripleO's Ansible after successful deployment
+	allNodesDataString, err := CopyFromPod(kclient, *pod, "openstackclient", "/home/cloud-admin/work/"+deployOpts.configVersion+"/playbooks/tripleo-ansible/group_vars/overcloud.json")
```
stuggi, line 239: I do not have an env right now, does this also have these details?
```
        "EndpointMap": {
            "parameter": "EndpointMapOverride",
        },
        "HostsEntry": {
            "parameter": "ExtraHostFileEntries",
        },
        "GlobalConfig": {
            "parameter": "GlobalConfigExtraMapData",
        },
```

api/v1beta1/openstackconfigversion_types.go
```diff
@@ -36,7 +36,7 @@ type OpenStackConfigVersionStatus struct {
 //+kubebuilder:subresource:status
 // +kubebuilder:resource:shortName=osconfigversion;osconfigversions
 // +operator-sdk:csv:customresourcedefinitions:displayName="OpenStack Config Version"
-// +kubebuilder:printcolumn:name="Generator",type="string",JSONPath=".spec.ConfigGeneratorName",description="Config Generator Name"
+// +kubebuilder:printcolumn:name="Generator",type="string",JSONPath=".spec.configGeneratorName",description="Config Generator Name"
```
dprince, line 39: Initially I had these things in the same patch. This fixes the name which needs to be camelcase. Not sure how I didn't copy that over to the 2nd review which already landed.

containers/agent/deploy.go
dprince: @stuggi just pushed a commit which splits out both filtered and non-filtered version of the file in the ConfigMap. Still need to add documentation on this to the README.
On the idea of making it a parameter while we could do that I suppose I'd rather not. Tripleclient has these as a constant so I'm not sure they change that often.

containers/agent/deploy.go
dprince: Well, it did last time I tested. These take a while to test as you have to wait for a full TripleO deploy to finish. Trying it again this morning.

dprince: yes, these are all in the resulting environment file

README.md
```diff
@@ -1545,6 +1545,25 @@ openstackbackuprestore   restore    openstackbackupsave-1641928378   Restored
...
 
+Exported stack data config map
+--------------------------------------------------------
+
+The OSP Director Operator automatically creates a ConfigMap after each OSDeploy resource finishes executing. This ConfigMap is named after the OSDeploy resource name and prefixed with tripleo-exports-. For example tripleo-exports-default would be the name of the ConfigMap for the 'default' OSDeploy resource. Each ConfigMap contains 2 YAML files:
+
+| Filename    | Description | TripleO Command Equivalent |
+| ----------- | ----------- | -------------------------- |
+| ctlplane-export.yaml | Used with multiple stacks for DCN | overcloud export |
+| ctlplane-export-filtered.yaml | Used for multiple stacks with Cells | overcloud cell export |
```
stuggi, line 1556: maybe we want to be more specific that this is used for cell stacks where the cell controllers get installed. if you split out again cell computes from a cell controller stack you'd use the ctlplane-export.yaml from the cell controller stack.

containers/agent/deploy.go
```diff
@@ -93,6 +191,111 @@ func ExecPodCommand(kclient kubernetes.Clientset, pod corev1.Pod, containerName
...
```
stuggi, line 242: not required in this patch, but we might also want to get the clouds.yaml and /etc/hosts from the deployment. in case there is an issue with the storage of the openstackclient, thats some useful information to restore it.

stuggi (commented):
just two minor comments, other then that it looks good to me.

containers/agent/deploy.go
dprince: @stuggi yeah, maybe we file this as Jira tickets and can follow up on it as a "core" issue.

README.md
dprince: Sure, I added a note to indicate the cell stacks as "Controllers"

stuggi (approved):
/lgtm

PR #521, opened 2022-02-24
[OSPK8-322] Introduce ipset again to also support service VIPs

stuggi:
This re-introduce the ipset concept, which gives the flexibility
to request separate IPs for additional services, which is required
for OSP17 (OVNDBs, Redis). With the IPset also used for the
other resources (osclient, vmset, bmset) all of them use the
same mechanism and workflow to create hostnames and assign IPs.

With this change also alligns the scale up/down steps/procedure
from vmset and bmset, as they were not alligned.


stuggi:
set to draft as I need to do some more testing tomorrow, but feel free to comment

controllers/openstackcontrolplane_controller.go
```diff
@@ -989,3 +813,112 @@ func (r *OpenStackControlPlaneReconciler) createOrUpdateOpenStackClient(
...
+		return ctrl.Result{}, err
+	}
+
+	var ctrlResult ctrl.Result
+	currentLabels := instance.DeepCopy().Labels
+	//
+	// add osnetcfg CR label reference which is used in the in the osnetcfg
+	// controller to watch this resource and reconcile
+	//
+	instance.Labels, ctrlResult, err = openstacknetconfig.AddOSNetConfigRefLabel(r, instance, cond, vipNetworksList[0])
```
abays, line 837: Could there ever be a need to loop through all networks in `vipNetworksList` here?  Just curious as to why the first network in the list and nothing else.

controllers/openstackipset_controller.go
```diff
@@ -0,0 +1,397 @@
...
+				cond.Message = "Failed to get list of all BareMetalHost(s)"
+				cond.Reason = ospdirectorv1beta1.BaremetalHostCondReasonListError
+				cond.Type = ospdirectorv1beta1.ConditionType(ospdirectorv1beta1.BaremetalSetCondTypeError)
+
+				return reconcile.Result{}, err
+			}
+
+			for _, bmh := range baremetalHostsList.Items {
+				hostStatus.HostRef = bmh.Name
+			}
```
abays, line 276: Are we only ever expecting one item in the list?  Just wondering if things would get wonky if somehow this list had more than one element.

controllers/openstacknetconfig_controller.go
```diff
@@ -819,14 +815,41 @@ func (r *OpenStackNetConfigReconciler) createOrUpdateOpenStackMACAddress(
...
 		client.MatchingLabels(labelSelector),
 	}
 
+	ipSetList := &ospdirectorv1beta1.OpenStackIPSetList{}
+	if err := r.GetClient().List(
+		context.Background(),
+		ipSetList,
+		listOpts...,
+	); err != nil && !k8s_errors.IsNotFound(err) {
+		return macAddress, err
```
abays, line 834: Would it be safer to return `nil` here instead of `macAddress`.  I guess it doesn't really matter, since the caller should see the non-nil `err` and act on that.

pkg/openstackipset/funcs.go
```diff
@@ -0,0 +1,244 @@
...
+	// get OSNetCfg object
+	//
+	osnetcfg := &ospdirectorv1beta1.OpenStackNetConfig{}
+	err = r.GetClient().Get(context.TODO(), types.NamespacedName{
+		Name:      strings.ToLower(obj.GetLabels()[openstacknetconfig.OpenStackNetConfigReconcileLabel]),
+		Namespace: obj.GetNamespace()},
+		osnetcfg)
+	if err != nil {
+		cond.Message = fmt.Sprintf("Failed to get %s %s ", osnetcfg.Kind, osnetcfg.Name)
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.MACCondReasonError)
```
abays, line 106: Is this `MAC` reason the intended reason for an OSNetConfig error?

pkg/openstackipset/funcs.go
```diff
@@ -0,0 +1,244 @@
...
+			r,
+			obj,
+			cond,
+			osnetcfg,
+			networks,
+			hostname,
+			&hostStatus,
+		)
+		if err != nil {
+			return status, reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
abays, line 124: Do we want to return `err` here since it's non-nil?

abays: I guess `openstacknetconfig.WaitOnIPsCreated` only ever returns "not found" errors, so perhaps it's unimportant.

pkg/openstackipset/funcs.go
```diff
@@ -0,0 +1,244 @@
...
+			err = common.WrapErrorForObject(cond.Message, obj, err)
+
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		cond.Message = fmt.Sprintf("Failed to create or update %s %s ", ipSet.Kind, ipSet.Name)
+		cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.OsClientCondReasonError)
```
abays, line 189: This might be the wrong reason?

pkg/openstackipset/funcs.go
```diff
@@ -0,0 +1,244 @@
...
+	if op != controllerutil.OperationResultNone {
+		cond.Message = fmt.Sprintf("%s - operation: %s", cond.Message, string(op))
+
+		common.LogForObject(
+			r,
+			cond.Message,
+			obj,
+		)
+	}
+	cond.Reason = ospdirectorv1beta1.ConditionReason(ospdirectorv1beta1.OsClientCondReasonCreated)
```
abays, line 206: Same here with regards to reason

abays (commented):
A few minor comments, but otherwise seems ready to me

controllers/openstackcontrolplane_controller.go
stuggi: no, unless we ever support having multiple osnetcfgs. This network is only used to identify the osnetcfg object.

controllers/openstackipset_controller.go
stuggi: yes, we expect that there is a 1:1 mapping for osp hostname and bmh. if we get multiple BMHs with that above labelselector, there was an error in the bmset provisioning of bmh hosts [1]. 

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstackbaremetalset_controller.go#L1025

controllers/openstacknetconfig_controller.go
stuggi: yes, it should not matter as we only check for non-nil error in the caller

pkg/openstackipset/funcs.go
stuggi: good spot, thanks. will fix that

pkg/openstackipset/funcs.go
stuggi: I wanted the controller just reconcile with the info message and not log an error. thats why I return a non default ctrl.Result and nil. if we ever return another error that would get logged as an error.

pkg/openstackipset/funcs.go
stuggi: good spot, thanks. will fix that

pkg/openstackipset/funcs.go
stuggi: good spot, thanks. will fix that

abays (approved):
/lgtm

PR #529, opened 2022-03-09
[OSPK8-447] Fix multi subnet network identification and routes rendering/template

stuggi:
- Use subnetName instead of networkNameLower to identify network

The correct way to identify the network is the subnetName because the networkNameLower can return multiple networks when multiple subnets are used.

- [OSPK8-447] fix network routes for IPv6 multi subnet
The routes parameter of the subnet in `network_data.yaml`, rendered by the operator, use `routes` instead of the expected `routes_ipv6`.

Also the routes parameter rendered in environments/network-environment.yaml which are named <netName>Routes get usually set on the neutron network host_routes property.

```
      {{network.name}}Routes:
          Routes for the {{network.name_lower}} network traffic.
          JSON route e.g. [{'destination':'10.0.0.0/16', 'nexthop':'10.0.0.1'}]
          Routes are added to the host_routes property on the subnet in neutron
          when the network and subnet is created.

```

Since there is no neutron we add the {{network.name}}Routes to the nic template and concat the two lists:
```
      routes:
        list_concat_unique:
          - get_param: {{ $net.Name }}Routes
          - get_param: {{ $net.Name }}InterfaceRoutes
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1227,6 +1227,30 @@ Update the `tarballConfigMap` configmap to add the `roles_data.yaml` file to the
 
 ## Create NIC templates for the new roles
 
+### Default network routes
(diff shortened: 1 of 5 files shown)


abays (approved):
/lgtm

PR #530, opened 2022-03-09
Support interrupting a deploy job

olliewalsh:
The tripleo-deploy.sh oc exec shell creates a new process group. Can use
this to terminate the deployment.

Set terminateGracePeriod to 60s for the deploy job container.
Add a signal handler goroutine to the deploy job that exec a new
tripleo-deploy-term.sh script on the openstackclient.
This script will send a SIGTERM to the process group and, if it has not
not terminated after ~30s~50s, send a SIGKILL.

### DIFF
--- a/templates/openstackclient/bin/tripleo-deploy.sh
+++ b/templates/openstackclient/bin/tripleo-deploy.sh
@@ -4,6 +4,18 @@ umask 022
 
 export PWD=/home/cloud-admin
 CONFIG_VERSION=${CONFIG_VERSION:?"Please set CONFIG_VERSION."}
+sudo bash -c 'mkdir -p /var/run/tripleo-deploy && chown '$(whoami)' /var/run/tripleo-deploy'
+RUNDIR="/var/run/tripleo-deploy/$CONFIG_VERSION"
+mkdir -p $RUNDIR
+PGIDFILE=$RUNDIR/pgid
+trap "rm -f $PGIDFILE" EXIT
+# Assume PGID==$PPID which is the case when run via oc exec
+# Alternatively could do something like:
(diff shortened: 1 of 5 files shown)


templates/openstackclient/bin/tripleo-deploy.sh
```diff
@@ -4,6 +4,18 @@ umask 022
 
 export PWD=/home/cloud-admin
 CONFIG_VERSION=${CONFIG_VERSION:?"Please set CONFIG_VERSION."}
+sudo bash -c 'mkdir -p /var/run/tripleo-deploy && chown '$(whoami)' /var/run/tripleo-deploy'
```
stuggi, line 7: could also use an emptyVol for /var/run/tripleo-deploy and change the owner in the init container, but I am also ok to have it as is

dprince: wait, if we use emptyDir for /var/run/tripleo-deploy won't that potentially get rid of the ansible log too?

dprince: wait, if we use emptyDir for /var/run/tripleo-deploy won't that potentially get rid of the ansible log too?

dprince: Oh, Nevermind. This is just the RUNDIR. Not workdir.

dprince: Oh, Nevermind. This is just the RUNDIR. Not workdir.

dprince:
Looks good to me. The /var/run/tripleo-deploy emptydir could come in a later patch

dprince:
Looks good to me. The /var/run/tripleo-deploy emptydir could come in a later patch

stuggi (approved):
/lgtm

PR #533, opened 2022-03-11
Update webhook/manifests.yaml for removed webhook

dprince:
In eab3f6d203e79590bb95a4a406c15363865cc071 we removed a duplicate
webhook. This updates the config/webhook/manifests.yaml for
the associated change.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -337,27 +337,6 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
-      path: /validate-osp-director-openstack-org-v1beta1-openstackprovisionserver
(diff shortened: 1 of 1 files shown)


dprince:
@rabi double check this

rabi:
I thought it was updated in the generated webhook manifest and included in the original patch. Not sure how it got dropped from the PR:/

dprince:
> I thought it was updated in the generated webhook manifest and included in the original patch. Not sure how it got dropped from the PR:/

It should still get updated when we build the operator but as we check in these files running 'make manifests' I think will update these files pre-commit.

abays (approved):
/lgtm

PR #536, opened 2022-03-15
Fix openstackclient pod update

olliewalsh:
Pod update logic only handled env vars. This adds support for any spec change.


controllers/openstackclient_controller.go
```diff
@@ -441,122 +496,85 @@ func (r *OpenStackClientReconciler) podCreateOrUpdate(
...
-				Env:             env,
-				VolumeMounts:    volumeMounts,
-			},
+	common.InitMap(&pod.Labels)
+	pod.Spec = corev1.PodSpec{}
+	pod.Spec.SecurityContext = &corev1.PodSecurityContext{}
+	pod.Spec.DNSConfig = &corev1.PodDNSConfig{}
+	pod.Spec.Containers = []corev1.Container{
+		{
+			Name: "openstackclient",
```
stuggi, line 513: this could probably be instance.name?

pkg/openstackclient/errors.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (e *ForbiddenPodSpecChange) Error() string {
+	return fmt.Sprintf("cannot update Pod spec field: %s", e.Field)
+}
+
+// IsForbiddenPodSpecChange -
+func IsForbiddenPodSpecChange(err error) bool {
+	return errors.Is(err, &ForbiddenPodSpecChange{}) || k8s_errors.IsInvalid(err)
+}
```
stuggi, line 38: could also add all this to common/pod.go as its generic and could be used in future for other pod handling?

controllers/openstackclient_controller.go
```diff
@@ -441,122 +496,85 @@ func (r *OpenStackClientReconciler) podCreateOrUpdate(
...
+		}
+		if len(instance.Spec.DNSSearchDomains) != 0 {
+			pod.Spec.DNSConfig.Searches = instance.Spec.DNSSearchDomains
+		} else {
+			pod.Spec.DNSConfig.Searches = []string{}
+		}
+
+		if isPodUpdate && len(pod.Status.InitContainerStatuses) > 0 {
+			if pod.Status.InitContainerStatuses[0].Image != pod.Spec.InitContainers[0].Image {
+				// init container image is mutable but does nothing so force a delete
```
stuggi, line 573: its the same also for the "main" container image, but since we use the same image for both its ok to handle this via the init container image

controllers/openstackclient_controller.go
```diff
@@ -441,122 +496,85 @@ func (r *OpenStackClientReconciler) podCreateOrUpdate(
...
+			pod.Spec.DNSConfig.Nameservers = []string{}
+		}
+		if len(instance.Spec.DNSSearchDomains) != 0 {
+			pod.Spec.DNSConfig.Searches = instance.Spec.DNSSearchDomains
+		} else {
+			pod.Spec.DNSConfig.Searches = []string{}
+		}
+
+		if isPodUpdate && len(pod.Status.InitContainerStatuses) > 0 {
+			if pod.Status.InitContainerStatuses[0].Image != pod.Spec.InitContainers[0].Image {
```
stuggi, line 572: just to mention another possibility, could also check this in the webhook where we have the old + new spec to see if it changed and set some status field.

stuggi (commented):
looks good to me, just one question if we should move the ForbiddenPodSpec handling to common/pod.go as it seems to be generic?

pkg/openstackclient/errors.go
olliewalsh: Or create common/errors.go?

controllers/openstackclient_controller.go
olliewalsh: changing the main container image works. In fact that is the only thing that can be changed in a pod update

stuggi: yes, but iirc it also does not auto restart the pod

pkg/openstackclient/errors.go
stuggi: either is fine for me, I just think it might be good in the common space

controllers/openstackclient_controller.go
olliewalsh: Correct, k8s auto restarts the main container (not the pod) when the main container image is changed. It does absolutely nothing when the init container image changes but we need to ensure init reruns as the new main container could depend on something from the new init image.

While I could just trigger a delete here when the main container image changes I've intentional avoided this. I'm trying to keep the logic generic as I expect we will copy this logic for other use cases that could be using different images for the init and main containers.

controllers/openstackclient_controller.go
olliewalsh: Could I guess  (I've not change it in this patch). Does it make much difference in practice?

pkg/openstackclient/errors.go
olliewalsh: Yes, I was thinking the same but I figured we could come back to it later if/when we add more error type (this is the first one, IIRC).

olliewalsh: I'll create common/errors.go as this seems to be the convention (e.g https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors)

controllers/openstackclient_controller.go
stuggi: no it does not make a different, and thinking about it again it might be even better to have it fixed, like when you create a tempest openstackclient, the container will sill be openstackclient, instead of tempest. probably easier for debugging

stuggi (approved):
/lgtm

PR #541, opened 2022-03-21
Handle DeployIdentifier

olliewalsh:
DeployIdentifier should be unique to every deploy invocation. As we
generate playbooks and re-use them we cannot do this via a heat
parameter. Instead set this to a marker string when generating playbooks
that can be replaced later at deploy time.

Also add a deploy CR param to skip setting this, as it can be used to
speed up scale-up deployments.


abays:
Does this depend on #534 ?

olliewalsh:
> Does this depend on #534 ?

No. This is about setting a require tripleo heat parameter that is used, for example, to force certain contain restarts on every deployment (it's passed as an env var to containers). It must be unique for every deployment invocation.

#534 proposes adding a similarly param to the deploy CR, it has similar name/behaviour but completely different context. It just prompted me to look at how we are setting the t-h-t param and realise the logic isn't correct.

stuggi (approved):
/lgtm

PR #542, opened 2022-03-21
Ensure correct images are used for provision server

olliewalsh:
The provision server should always use the images specified in the
bundle CSV.
With the current approach of setting images in the spec and defaulting
in the webhook, users are likely to overlook updating the CRs after
updating the operator.
This could lead to mis-matched operator and agent images being used
together, or missing out on critial security fixes in the apache image.

This removes the images from the provision server spec completely and
ensures the correct images are always used in the Reconcile function.

### DIFF
--- a/controllers/openstackprovisionserver_controller.go
+++ b/controllers/openstackprovisionserver_controller.go
@@ -55,9 +55,10 @@ var (
 // OpenStackProvisionServerReconciler reconciles a ProvisionServer object
 type OpenStackProvisionServerReconciler struct {
 	client.Client
-	Kclient kubernetes.Interface
-	Log     logr.Logger
-	Scheme  *runtime.Scheme
+	Kclient  kubernetes.Interface
+	Log      logr.Logger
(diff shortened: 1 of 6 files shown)


controllers/openstackprovisionserver_controller.go
```diff
@@ -402,7 +403,7 @@ func (r *OpenStackProvisionServerReconciler) deploymentCreateOrUpdate(
 			Containers: []corev1.Container{
 				{
 					Name:            "osp-httpd",
-					Image:           instance.Spec.ApacheImageURL,
+					Image:           r.Defaults.ApacheImageURL,
```
pinikomarov, line 406: Hi Oli, question about the reconcile here :
So if the contoller finds that an image is is not the right one , what will happen here, 
it will delete the old one and recreate it with the updated Image version ?

pinikomarov: ack

PR #543, opened 2022-03-22
Replicate 16.2 parameter merge behaviour

olliewalsh:
In a regular OSP16 deployment, parameters are merged by heat-client if
they are a map type.
This does not happen automatically in OSP17 regular deployments and
Director operator (for both 16.x and 17.x) which use ephemeral heat.
The merge strategy must be set explicitly in each environment file.

To workaround this, and remain compatible with existing 16.x
environment files and 3rd party integrations, automatically inject the
merge strategy for all map type parameters if not explicitly set.

### DIFF
--- a/templates/openstackconfiggenerator/bin/create-playbooks.sh
+++ b/templates/openstackconfiggenerator/bin/create-playbooks.sh
@@ -85,14 +85,13 @@ openstack tripleo container image prepare $PREPARE_ENV_ARGS -r roles_data.yaml -
 mkdir -p $HOME/tripleo-deploy
 rm -rf $HOME/tripleo-deploy/overcloud-ansible*
 
-time openstack stack create --wait \
+HEAT_ENVIRONMENT_FILES="
     -e $TEMPLATES_DIR/overcloud-resource-registry-puppet.yaml \
     -e $TEMPLATES_DIR/tripleo-overcloud-images.yaml \
     -e $TEMPLATES_DIR/environments/deployed-server-environment.yaml \
     -e $TEMPLATES_DIR/environments/docker-ha.yaml \
     -e $TEMPLATES_DIR/environments/network-isolation.yaml \
     -e $TEMPLATES_DIR/environments/network-environment.yaml \
-    -e ~/config-passwords/tripleo-overcloud-passwords.yaml \
 {{- range $i, $value := .TripleoEnvironmentFiles }}
     -e $TEMPLATES_DIR/{{ $value }} \
 {{- end }}
@@ -102,6 +101,17 @@ time openstack stack create --wait \
 {{- range $key, $value := .TripleoCustomDeployFiles }}
     -e {{ $key }} \
 {{- end }}
+"
+
+{{- if eq .OSPVersion "16.2" }}
+# Replicate the (undocumented) heat client merging of map parameters
+# STF, Contrail, Trilio and presumably others currently rely on it
+/home/cloud-admin/process-heat-environment.py ${HEAT_ENVIRONMENT_FILES}
+{{- end }}
+
+time openstack stack create --wait \
+    -e ~/config-passwords/tripleo-overcloud-passwords.yaml \
+    ${HEAT_ENVIRONMENT_FILES} \
     -t overcloud.yaml overcloud
 
 mkdir -p $HOME/ansible
--- a/pkg/openstackconfiggenerator/volumes.go
+++ b/pkg/openstackconfiggenerator/volumes.go
@@ -47,6 +47,12 @@ func GetVolumeMounts(instance *ospdirectorv1beta1.OpenStackConfigGenerator) []co
 			ReadOnly:  true,
 		},
 		{
+			Name:      "openstackconfig-scripts",
+			MountPath: "/home/cloud-admin/process-heat-environment.py",
+			SubPath:   "process-heat-environment.py",
+			ReadOnly:  true,
+		},
+		{
 			Name:      "git-ssh-config",
 			MountPath: "/mnt/ssh-config/git_id_rsa",
 			SubPath:   "git_id_rsa",
@@ -116,6 +122,10 @@ func GetVolumes(instance *ospdirectorv1beta1.OpenStackConfigGenerator) []corev1.
 						{
 							Key:  "create-playbooks.sh",
 							Path: "create-playbooks.sh",
+						},
+						{
+							Key:  "process-heat-environment.py",
+							Path: "process-heat-environment.py",
 						},
 					},
 				},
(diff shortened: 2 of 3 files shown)


templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
```diff
@@ -28,6 +28,17 @@ resource_registry:
 {{- end }}
 {{- end }}
 
+# Parameter merge strategies
+parameter_merge_strategies:
```
olliewalsh, line 32: @rabi could we just set the default to merge for 16.x?

olliewalsh: Looking at the heat merge code that wouldn't work - it would join string params and add number params!

rabi: I don't think it's required for 16/17. It's not documented that maps would be merged in anyway. In case someone one wants to change these parameters, they are expected to provide full ServiceNetMap, EndpointMap[1] etc and the ones in the last environment would override.

However, we've changed some of those in master where we specify merge strategies[2]. parameter_merge_strategy: merge for a map/list won't do a deep merge (as you mentioned), just a simple merge[3]. 

[1] https://github.com/openstack/tripleo-heat-templates/blob/stable/train/environments/ssl/tls-endpoints-public-dns.yaml#L14 
[2] https://github.com/openstack/tripleo-heat-templates/blob/master/overcloud-resource-registry-puppet.j2.yaml#L345
[3] https://github.com/openstack/heat/blob/master/heat/common/environment_util.py#L58

olliewalsh: STF is relying on this undocumented merge behaviour for ExtraConfig

rabi: I've not seen any issues reported for OSP17 yet though.  I guess would expect them to change instead and not depend on undocumented behavior though:)

olliewalsh: Probably just not hit this for 17 yet :-)

E.g this is relying on ServiceNetMap merging: https://github.com/tungstenfabric/tf-tripleo-heat-templates/blob/stable/train/environments/contrail/contrail-services.yaml#L2
And for STF the docs say to include the qdr-edge-only.yaml environment file (which sets ExtraConfig) and then creating a custom environment file that also sets ExtraConfig.

olliewalsh: Trilio too https://github.com/trilioData/triliovault-cfg-scripts/blob/master/redhat-director-scripts/rhosp16.2/environments/trilio_env.yaml#L17

rabi: That would be a STF bug (not TripleO), if reported ;) I only changed in xena[1] in THT. Probably we should not change something that's not inline documentation(for 17). 

[1] https://github.com/openstack/tripleo-heat-templates/commit/b253d564f7cccd0a818239a3c877229242acfe55

templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
```diff
@@ -28,6 +28,17 @@ resource_registry:
...
+  ServiceNetMap: merge
+  VipSubnetMap: merge
+  EndpointMap: merge
+  ExtraConfig: merge
+{{- range $roleid, $role := .RolesMap }}
```
bogdando, line 37: roleid is unused, should it be _?

olliewalsh: same for all of the other loops in that file so I kept it consistent

templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
rabi: I don't see any difference in 16.2/17 for director-operator, as we're using stack api[1] and not overcloud deploy (which merges environment on the client side for 16.2). Not sure what changing for 16.2 only would achieve. 

Also, as I mentioned earlier just adding here won't help. It has to be added to all environment files where these parameters are mentioned.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/openstackconfiggenerator/bin/create-playbooks.sh#L88

rabi: Yes it's per file and for a reason[1].

"You can provide a default merge strategy and/or parameter specific merge strategies per environment file"

[1] https://docs.openstack.org/heat/latest/template_guide/environment.html#environment-merging

olliewalsh:
Looks like heat cannot handle this for us so falling back on client-side processing of the environment files. Seems to work ok:


Set ExtraConfig in 2 environment files:

```
cloud-names.yaml:
...
  ExtraConfig:
    foo: 1
    bar: 2
    baz: 3

updates.yaml:
...
  ExtraConfig:
    foo: 10
    def: 100
```

Pre-processing sets the merge strategy in both env files (the copies of them in the working dir):

```
parameter_merge_strategies:
  ExtraConfig: merge
```

This results in correctly merged extraconfig in the generated ansible:

```
group_vars/Controller:
extraconfig:
  bar: 2
  baz: 3
  def: 100
  foo: 10
```

templates/openstackconfiggenerator/bin/create-playbooks.sh
```diff
@@ -85,6 +85,28 @@ openstack tripleo container image prepare $PREPARE_ENV_ARGS -r roles_data.yaml -
...
+    $TEMPLATES_DIR/environments/network-environment.yaml \
+{{- range $i, $value := .TripleoEnvironmentFiles }}
+    $TEMPLATES_DIR/{{ $value }} \
+{{- end }}
+{{- range $key, $value := .TripleoDeployFiles }}
+    {{ $key }} \
+{{- end }}
+{{- range $key, $value := .TripleoCustomDeployFiles }}
+    {{ $key }} \
+{{- end }}
```
stuggi, line 106: looks good to me, just one clarification, should this environment list be always that same as in the stack create (except the password file)? Just thinking if it would be good if the script also uses -e for any parameter file and then  just create one var containing all env files and use it for both process-heat-env and stack create, like  PREPARE_ENV_ARGS. So that its only required to do any change in a single place.

olliewalsh: yeah, could do that

stuggi (approved):
/lgtm

PR #544, opened 2022-03-22
Change from CreateOrUpdate to CreateOrPatch and introduce ctx parameters

stuggi:
This changes to create/update resources from CreateOrUpdate to
CreateOrPatch. CreateOrPatch only performs the patch if the before
and after resources (minus status) differ. This should reduce the
amount of updates performed on the CR objects.
    
[1] https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.7/pkg/controller/controllerutil/controllerutil.go#L232

Also adds context parameter to methods/funcs where used and passes
the ctx from main reconcile method.

### DIFF
--- a/controllers/openstackbaremetalset_controller.go
+++ b/controllers/openstackbaremetalset_controller.go
@@ -500,7 +500,7 @@ func (r *OpenStackBaremetalSetReconciler) provisionServerCreateOrUpdate(
 			},
 		}
 
-		op, err := controllerutil.CreateOrUpdate(ctx, r.Client, provisionServer, func() error {
+		op, err := controllerutil.CreateOrPatch(ctx, r.Client, provisionServer, func() error {
 			// Assign the prov server its existing port if this is an update, otherwise pick a new one
 			// based on what is available
 			err := provisionServer.AssignProvisionServerPort(
--- a/controllers/openstackbaremetalset_controller.go
+++ b/controllers/openstackbaremetalset_controller.go
@@ -99,7 +99,7 @@ func (r *OpenStackBaremetalSetReconciler) Reconcile(ctx context.Context, req ctr
 
 	// Fetch the instance
 	instance := &ospdirectorv1beta1.OpenStackBaremetalSet{}
-	err := r.Client.Get(context.TODO(), req.NamespacedName, instance)
+	err := r.Get(ctx, req.NamespacedName, instance)
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile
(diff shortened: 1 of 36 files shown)


controllers/openstackdeploy_controller.go
```diff
@@ -120,20 +120,15 @@ func (r *OpenStackDeployReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		)
 
 		if statusChanged() {
-			if updateErr := r.Client.Status().Update(context.Background(), instance); updateErr != nil {
-				if err == nil {
-					err = common.WrapErrorForObject("Update Status", instance, updateErr)
-				} else {
-					common.LogErrorForObject(r, updateErr, "Update status", instance)
-				}
+			if updateErr := r.Status().Update(context.Background(), instance); updateErr != nil {
```
olliewalsh, line 123: should this use the existing context?

controllers/openstackbaremetalset_controller.go
```diff
@@ -872,7 +888,7 @@ func (r *OpenStackBaremetalSetReconciler) baremetalHostProvision(
 				instance.Status.BaremetalHosts[bmhStatus.Hostname] = bmhStatus
 
 				// update status with host assignment
-				if err := r.Client.Status().Update(context.TODO(), instance); err != nil {
+				if err := r.Status().Update(context.Background(), instance); err != nil {
```
abays, line 891: Curious as to why we're using `Background()` here instead of the passed-in `ctx`

controllers/openstackbaremetalset_controller.go
```diff
@@ -1489,7 +1513,7 @@ func (r *OpenStackBaremetalSetReconciler) checkBMHsAnnotatedForDeletion(
 			instance,
 		)
 
-		err = r.Client.Status().Update(context.TODO(), instance)
+		err = r.Status().Update(context.Background(), instance)
```
abays, line 1516: Same `Background()` vs `ctx` question here

controllers/openstackcontrolplane_controller.go
```diff
@@ -360,7 +366,7 @@ func (r *OpenStackControlPlaneReconciler) SetupWithManager(mgr ctrl.Manager) err
 			listOpts := []client.ListOption{
 				client.InNamespace(obj.GetNamespace()),
 			}
-			if err := r.Client.List(context.Background(), crs, listOpts...); err != nil {
+			if err := r.List(context.Background(), crs, listOpts...); err != nil {
```
abays, line 369: Should this use `ctx`?

abays: I misread this one, my bad

controllers/openstackbaremetalset_controller.go
stuggi: as discussed on slack, I just wanted to make sure in some places to use a context which It is never canceled, but since we only support a single instance of the operator it probably does not make a different right now.

abays (approved):
/lgtm

PR #545, opened 2022-03-23
Upgrade to operator-sdk 1.14.0

dprince:
OSPK8-455 #close


stuggi:
> OSPK8-438 #close

should be OSPK8-455, right?

PR #551, opened 2022-03-30
[OSPK8-457] Allow adding new osnetattach definitions

stuggi:
Adding/updating an osnetcfg to add a new attachconfig fails with
"denied the request: bridge names may not be changed".
If there is a new attachcfg added to spec.attachConfigurations
we fail to find it in the oldInstance and therefore return with
an error like the bridge name probably changed.
Instead if the attachRef is in the spec of the old CR instance
* check if bridge names did not change
* otherwise expect it to be a new attachconfiguration/interface
  to configure on the workers.


abays (approved):
/lgtm

PR #552, opened 2022-03-30
[OSPK8-458] Do not wait in osnetcfg for nncp

stuggi:
If a deployment finished e.g. tempest runs via an openstackclient
resource. If this resource gets created when any of an NNCP target
nodes is not in configured state, nmstate is processing those and
the osnetcfg controller right now waits for this to be finished
before it assigns IP for the new openstackclient resource.

This patch removes this wait condition in osnetcfg and instead
proceed with the other resources managed by the controller.
Instead a validation got added to the osnetdeploy webhook to stop
the OSP deployment, if any nncp is not in configured state. Also
gets the condition message of the nncps related
NodeNetworkConfigurationEnactment to point to where the issue is.
It also introduce a SkipNNCPValidation parameter to overwrite this
if needed.


api/v1beta1/openstackdeploy_webhook.go
```diff
@@ -52,3 +65,142 @@ func (r *OpenStackDeploy) Default() {
...
+						condition.Message,
+						nnce.Name,
+						nnceError,
+					)
+
+					// if user skip validation, just log the failed validation
+					if r.Spec.SkipNNCPValidation {
+						openstackdeploylog.Info(fmt.Sprintf("SkipNNCPValidation configured! %s", msg))
+					} else {
+						info := "It's recommended to resolve the failures before proceeding with the deployment. If this validation should be skipped add spec.SkipNNCPValidationvalidate: true"
```
abays, line 192: s/SkipNNCPValidationvalidate/SkipNNCPValidation

abays (commented):
One small thing but otherwise +1

api/v1beta1/openstackdeploy_webhook.go
stuggi: done, also added it to the check to ValidateUpdate

api/v1beta1/openstackdeploy_webhook.go
```diff
@@ -52,3 +65,149 @@ func (r *OpenStackDeploy) Default() {
...
+						),
+					}
+
+					if err := webhookClient.List(context.TODO(), nnceList, listOpts...); err != nil {
+						return err
+					}
+
+					var nnceError string
+					var nnce nmstatev1alpha1.NodeNetworkConfigurationEnactment
+					for _, nnce = range nnceList.Items {
```
bogdando, line 174: Did you miss an empty list check, like you did above?

/late-to-the-party

bogdando:
Only issuing a warning would not provide the best UX. I would expect the failed nodes to be placed as pending to be configured in some future reconciliation loops additionally to the warning.

Ultimately, if skip nncp is set, deployment should not start until that pending list cleared out, IMHO
or the deployment could be repeated  by the controller for that list of members until it is cleared out eventually

PR #554, opened 2022-03-30
Updates for SDK 1.17 (requires golang 1.17)

dprince:
OSPK8-461 #close


dprince:
This will require us to bump our downstream CI to golang 1.17. Need to check on how to do this so we can continue to have golang 1.16 on the stable branch.

Makefile
```diff
@@ -1,22 +1,55 @@
...
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
-# CI tools repo for running tests
-CI_TOOLS_REPO := https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci
+# IMAGE_TAG_BASE defines the docker.io namespace and part of the image name for remote images.
+# This variable is used to construct full image tags for bundle and catalog images.
+#
+# For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
+# my.domain/foo-bundle:$VERSION and my.domain/foo-catalog:$VERSION.
+IMAGE_TAG_BASE ?= my.domain/foo
```
abays, line 32: Since this isn't commented-out, should we make this something specific to our project?

olliewalsh:
> This will require us to bump our downstream CI to golang 1.17. Need to check on how to do this so we can continue to have golang 1.16 on the stable branch. @olliewalsh

Do you mean for the buildah jenkins job? Might be better to change the job to fetch the version from Dockerfile

config/webhook/manifests.yaml
```diff
@@ -8,6 +7,7 @@ metadata:
 webhooks:
 - admissionReviewVersions:
   - v1
+  - v1beta1
```
abays, line 10: I'm curious as to why some of the webhooks here have both `v1` and `v1beta1` for `admissionReviewVersions` while others only have `v1`.  Is this because some have been around longer and need to be backwards compatible, whereas others are newer?  It looks like it varies for both mutating and validating webhooks.

stuggi: check https://github.com/openstack-k8s-operators/osp-director-operator/pull/555

abays: It would seem that #555 has only `v1` for everything.  That's why I'm a bit puzzled by the inclusion of both `v1` and `v1beta1` in places here.

stuggi: yes, thats from the changes we landed in parallel which added new webhooks. we should probably first land https://github.com/openstack-k8s-operators/osp-director-operator/pull/555 and then this here

dprince:
> > This will require us to bump our downstream CI to golang 1.17. Need to check on how to do this so we can continue to have golang 1.16 on the stable branch. @olliewalsh
> 
> @dprince Do you mean for the buildah jenkins job? Might be better to change the job to fetch the version from Dockerfile

Yes, I'm looking into it today.

config/webhook/manifests.yaml
dprince: Exactly. I rebased this on Martin's fix so we should be good

Makefile
dprince: Yeah, I was torn on setting this. It gets used to construct the BUNDLE_IMG and CATALOG_IMG which we typically explicitly set anyway in our scripts so it isn't required.

I guess we could set it to IMAGE_TAG_BASE ?= quay.io/osp-director-operator/osp-director-operator but I thought someone might inadvertently build and push to our default registry by mistake. WDYT?

Could add a comment to this regard though @abays

Makefile
```diff
@@ -98,63 +105,125 @@ golangci: get-ci-tools
...
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+##@ Development
+
+.PHONY: manifests
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
stuggi, line 130: since you now switch it back to `config/crd/bases` , we could remove /conf dir with the CRDs again?

dprince: @stuggi yes, and I now see where I messed things up. Initially I did all the operator-sdk upgrades in one big branch. Then I split it into two patches to make it easier to review. Thus I didn't catch this :/

.github/workflows/build-osp-director-operator.yaml
```diff
@@ -160,7 +160,7 @@ jobs:
       uses: redhat-actions/openshift-tools-installer@v1
       with:
         source: github
-        operator-sdk: '1.13.1'
+        operator-sdk: '1.18.2'
```
olliewalsh, line 163: Also need to install go here (see https://github.com/openstack-k8s-operators/osp-director-operator/pull/558/commits/b3cdc7a648bb55b31423cf8eade64035fea43af6)

Maybe we could use this to pull the version from go.mod - https://github.com/marketplace/actions/go-version-action

olliewalsh (changes requested):

.github/workflows/build-osp-director-operator.yaml
olliewalsh: FYI if you want to test these changes - a while back I added logic to these github actions so that they work with our forks.

Just need to add credentials in https://github.com/dprince/osp-director-operator/settings/secrets/actions:
QUAY_USERNAME
QUAY_PASSWORD
REDHATIO_USERNAME
REDHATIO_PASSWORD

dprince: @olliewalsh thanks. I rebased this morning. Should be good now. I will configure the secrets...

stuggi (commented):
looks good to me to go if others are ok as well

PR #555, opened 2022-03-31
some fixes after 1.14 upgrade

stuggi:


conf/osp-director.openstack.org_openstackbackuprequests.yaml
```diff
@@ -0,0 +1,147 @@
+
```
stuggi, line 1: conf is the new CRD dir from https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/Makefile#L79

config/manager/controller_manager_config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
```
stuggi, line 1: this was missing probably due to gitignore

api/v1beta1/openstackclient_webhook.go
```diff
@@ -94,7 +94,7 @@ func (r *OpenStackClient) Default() {
 
 }
 
-// +kubebuilder:webhook:verbs=create;update;delete,path=/validate-osp-director-openstack-org-v1beta1-openstackclient,mutating=false,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackclients,versions=v1beta1,name=vopenstackclient.kb.io,admissionReviewVersions={v1,v1beta1}
+// +kubebuilder:webhook:verbs=create;update;delete,path=/validate-osp-director-openstack-org-v1beta1-openstackclient,mutating=false,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackclients,versions=v1beta1,name=vopenstackclient.kb.io,admissionReviewVersions=v1
 
 var _ webhook.Validator = &OpenStackIPSet{}
```
abays, line 99: Not that this PR added this, but it seems that a lot of these `<x>_webhook.go` files have `var _ webhook.Validator = &OpenStackIPSet{}` instead of the CRD struct that actually pertains to them

stuggi: good catch!! I'll fix that as part of the PR I submitting later today

abays (approved):
/lgtm

PR #556, opened 2022-03-31
[OSPK8-417] Get DNS settings from osnetcfg if not specified

stuggi:
domainName, dnsSearchDomains and dnsServers are already set on the
osnetcfg object. Lets use those for ctlplane, vmset, bmset and osclient
if not manually specified in the CR of the relevant CRD.

Also fixes an issue in ipset where nil error can happen on status.hosts not being initialized.


api/v1beta1/openstackbaremetalset_webhook.go
```diff
@@ -57,6 +57,17 @@ func (r *OpenStackBaremetalSet) ValidateCreate() error {
...
+	//
+	// Fail early on create if osnetcfg ist not found
+	//
+	_, err := GetOsNetCfg(webhookClient, r.GetNamespace(), r.GetLabels()[OpenStackNetConfigReconcileLabel])
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("error getting OpenStackNetConfig %s - %s: %s",
+			r.GetLabels()[OpenStackNetConfigReconcileLabel],
+			r.Name,
+			err))
+	}
```
abays, line 69: Do we require the user to create an `OSBaremetalSet` CR with this label already present?  I'm just trying to understand how the `OpenStackNetConfigReconcileLabel` label would even be present when the user does `oc apply/create` on the CR YAML they've written.

abays:
One question from me but otherwise +1

api/v1beta1/openstackbaremetalset_webhook.go
stuggi: no, the Default webhook runs before the validation [1] and as you can see in the first Default task bellow we add the
OpenStackNetConfigReconcileLabel label based on the provided network provided in the CR. This one here just
makes sure to stop early and return the info to the user, instead of just log it in the operator log (as in the Defaulter).

[1] https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks

abays: Ack

abays (approved):
/lgtm

PR #558, opened 2022-04-01
[OSPK8-460] Replicate StackAction behaviour for 16.2

olliewalsh:
Pacemaker restart, Octavia, and tripleo-network-config rely on the
stack_action (CREATE or UPDATE) ansible fact or hieradata.
To ensure this works correctly with director operator's ephemeral
heat we need to replicate the tripleoclient logic.
As we do not have the existance of a heat stack to rely on we need a
different approach.
Instead the existance of /var/lib/tripleo-config on any overcloud node
is used to determine the appropriate stack_action setting. This is
passed on the ansible-playbook command line to override the default
stack_action in global_vars (which is always CREATE). The resulting
hieradata is also updated.

### DIFF
--- a/templates/openstackclient/bin/tripleo-deploy.sh
+++ b/templates/openstackclient/bin/tripleo-deploy.sh
@@ -4,6 +4,7 @@ umask 022
 
 export PWD=/home/cloud-admin
 CONFIG_VERSION=${CONFIG_VERSION:?"Please set CONFIG_VERSION."}
+OSP_VERSION=${OSP_VERSION:?"Please set OSP_VERSION."}
 sudo bash -c 'mkdir -p /var/run/tripleo-deploy && chown '$(whoami)' /var/run/tripleo-deploy'
 RUNDIR="/var/run/tripleo-deploy/$CONFIG_VERSION"
 mkdir -p $RUNDIR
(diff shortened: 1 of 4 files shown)


stuggi (commented):
just thinking, wondering if we should introduce some pre-tasks for the osconfiggenerator, which we run via the agent container (similar to deploy) and then have the required information in the osconfiggenerator when it renders the tripleo-deploy-config [1]

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml#L33

olliewalsh:
> just thinking, wondering if we should introduce some pre-tasks for the osconfiggenerator, which we run via the agent container (similar to deploy) and then have the required information in the osconfiggenerator when it renders the tripleo-deploy-config [1]
> 
> [1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml#L33

What if the user deploys the generated playbooks more than once?

stuggi:
> > just thinking, wondering if we should introduce some pre-tasks for the osconfiggenerator, which we run via the agent container (similar to deploy) and then have the required information in the osconfiggenerator when it renders the tripleo-deploy-config [1]
> > [1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml#L33
> 
> What if the user deploys the generated playbooks more than once?

right, have not thought about this scenario. If we think there is a need for more general pre tasks for osdeploy and those to be handled on the operator side, we can do it in an osdeploy init container task and have the playbook as templates for it.

templates/openstackclient/bin/tripleo-deploy.sh
```diff
@@ -94,6 +95,41 @@ play() {
...
+      delegate_facts: yes
+    - copy:
+        dest: stack_action_override.yaml
+        content: |
+          stack_action: {{ 'UPDATE' if hostvars['localhost'].deploy_exists|default('0')|int > 0 else 'CREATE' }}
+      delegate_to: localhost
+      run_once: yes
+EOF`}}
+
+    ansible-playbook \
```
olliewalsh, line 127: TODO: the initial deploy could be split (e.g --limit Controller for the 1st deploy then --limit Compute for the 2nd) so safer to run with the same --limit args as the deploy we are about to run

templates/openstackclient/bin/tripleo-deploy.sh
```diff
@@ -120,11 +121,48 @@ play() {
...
+    # rely on it.
+    # As we can't use the existance of a heat stack, instead use the existance of /var/lib/tripleo-config on any
+    # overcloud host (created by the tripleo-bootstrap role) to determine if this is a "CREATE" or an "UPDATE"
+    cat <<EOF > set_stack_action_playbook.yaml
+{{` ---
+- hosts: overcloud
+  name: Determine the correct stack action for 16.2 deployments
+  gather_facts: false
+  tasks:
+    - stat:
```
r3d3mpt10n, line 136: So, we check any node on the overcloud for the existence of this directory, if it exists we set the stack_action: UPDATE. 

I wonder if this would cause issues for nodes being added to the stack? In that case, Heat's StackAction would be set to CREATE. So I wonder if forcing it to UPDATE globally would cause issues for those nodes? Off the top of my head, I think it should be fine, since we primarily use it to do specific things on the Controller nodes, the Computes should be ok.

Looking at the network-config component, it should be fine if the stack_action is UPDATE since we will also check to see if it has been run before, and if not we will run os-net-config:

https://github.com/openstack/tripleo-ansible/blob/stable/train/tripleo_ansible/roles/tripleo-network-config/tasks/main.yml#L113-L128

bshephar, line 136: So, we check any node in the overcloud group, and if the directory exists, then we set the stack_action to UPDATE for all nodes.

I just wonder if this will impact nodes during a scale out, since the Heat StackAction for those new nodes would be CREATE. Off the top of my head, I think this would be fine since the main logic around stack_action is used to manage things that exist on the Controller nodes. Computes should be ok, but it might be something worth testing to double check.

Looking at the network_config tasks, we have logic in there that should be fine even if the stack_action is set to UPDATE. It will run os-net-config if it can't find evidence of it being executed previously:

https://github.com/openstack/tripleo-ansible/blob/stable/train/tripleo_ansible/roles/tripleo-network-config/tasks/main.yml#L113-L128

olliewalsh: > I just wonder if this will impact nodes during a scale out, since the Heat StackAction for those new nodes would be CREATE

AFAICT for a regular scale-out or node replacement deployment StackAction would be UPDATE:
https://github.com/openstack/tripleo-common/blob/stable/train/tripleo_common/actions/deployment.py#L163

The logic here should give the same result under normal circumstances.

I think the case where the logic could differ is if a scale-out deploy limits the playbook to just the new nodes. However doing this is not strictly valid as it would not update /etc/hosts or ssh_known_hosts on the other compute nodes which is required for VM migration.

> Looking at the network_config tasks, we have logic in there that should be fine even if the stack_action is set to UPDATE.
Yes, I noticed that when looking for usage of stack_action. I did consider always setting stack_action to UPDATE as this would work for the network_config tasks. If the Octavia change could be backported I think it would work (the other related change for pacemaker restart has already been backported to train). However I think the approach here is more flexible as any existing deployments will not have pacemaker restart change.

bshephar: Ah yeah, you're right. 

In that case, I think this should be fine then. And only required for 16.2 which you're already checking for. Hopefully 17 makes life easier for everyone when RHOSP catches up with you osp-director-operator to use ephemeral heat. :)

stuggi (approved):
/lgtm

templates/openstackclient/bin/tripleo-deploy.sh
olliewalsh: ack, thanks Brendan!

PR #561, opened 2022-04-08
Fixes to allow the operator to run local

stuggi:
- Don't rely on webhookClient in common functions to run local
  
Right now the webclient is used in a couple of places, which works fine if the operator is deployed in the env, but fails
when running local with webhooks disabled. This changes the functions which uses them and instead pass the client.Client from the caller.

- Add METRICS_BIND_ADDRESS to Makefile allows operator to run local
  
If port 8080 is already in use to run the operator local it is required to specify a different ports for metrics to bind on. This adds METRICS_BIND_ADDRESS env var to control the address/port metrics to bind to. This allows the operator to be run local like:

`WATCH_NAMESPACE=openstack,openshift-machine-api,openshift-sriov-network-operator ENABLE_WEBHOOKS=false METRICS_BIND_ADDRESS=:8085 make run`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ endif
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
+# Metrics bind address
+METRICS_BIND_ADDRESS ?= :8080
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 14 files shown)


abays (approved):
/lgtm

Makefile
```diff
@@ -153,7 +155,7 @@ build: generate fmt vet ## Build manager binary.
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
-	go run ./main.go
+	go run ./main.go -metrics-bind-address ${METRICS_BIND_ADDRESS}
```
dprince, line 158: Changes like this to the Makefile may make future rebases for operator-sdk more difficult. Customizing the Makefile is still fine of course I'm starting to try to keep all our customizations in a standard section. 

Wondering if looking for the ENV variable in main.go might end up being cleaner.

PR #563, opened 2022-04-08
[OSPK8-430]Update kubevirt api and VMset disk restructure

stuggi:
- Update to use kubevirt api v0.51.0 and cdi api v1.42.0

- [OSPK8-430] Allow adding additional disks to a vm role
    Restructures disk specification for vmset/ctlplane VM roles.
    Adds OpenStackVMSetDisk type which describes any disk
    configured for the VM.
    Root of vmset now has RootDisk of type OpenStackVMSetDisk
    and AdditionalDisks of type []OpenStackVMSetDisk.
    OpenStackVMSetDisk with the parameters:
        - Name
        - DiskSize
        - StorageClass
        - StorageAccessMode
        - StorageVolumeMode
        - DedicatedIOThread

    Also adds IOThreadsPolicy and BlockMultiQueue to the root vmset spec:

    * DedicatedIOThread - Disks with dedicatedIOThread set to true will be
    allocated an exclusive thread. This is generally useful if a specific
    Disk is expected to have heavy I/O traffic, e.g. a database spindle.

    * If IOThreadsPolicy is default, use of IOThreads will be disabled. However,
    if any disk requests a dedicated IOThread, ioThreadsPolicy will be enabled
    and default to shared.
    When ioThreadsPolicy is set to auto IOThreads will also be "isolated" from
    the vCPUs and placed on the same physical CPU as the QEMU emulator thread.
    An ioThreadsPolicy of shared indicates that KubeVirt should use one thread
    that will be shared by all disk devices.

    * Block Multi-Queue is a framework for the Linux block layer that maps Device
    I/O queries to multiple queues.
    This splits I/O processing up across multiple threads, and therefor multiple
    CPUs. libvirt recommends that the number of queues used should match the
    number of CPUs allocated for optimal performance.

- [OSPK8-430] Disk validations via webhooks
    On CR create:
    * additional disks in a vm role
        - validate Disk Name is not 'rootdisk' as this is reserved for the boot disk
        - validate that disk name is uniq within the VM role
    
    On CR update:
    * rootdisk
        - validate DiskSize don't change
        - validate StorageAccessMode don't change
        - validate StorageClass don't change
        - validate StorageVolumeMode don't change
    
    * additional disks in a vm role
        - all check same update checks of the rootdisk
        - validate Disk Name is not 'rootdisk' as this is reserved for the boot disk
        - validate that disk name is uniq within the VM role


stuggi:
close in favor of https://github.com/openstack-k8s-operators/osp-director-operator/pull/569

PR #564, opened 2022-04-11
Fix agent and downloader image build

stuggi:
[1] removed the DOCKERFILE and DOCKER_BUILD_DIR env variables
from the docker-build target. As a result for the anget and
the downloader image also the operator image got build and
pushed.
To keep the Makefile as default as possible, this adds new
specific targets, osp-director-operator-agent-image-build and
osp-director-downloader-image-build to build these images.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/commit/efc90f909d9d62928acd33dcf57163d895525869#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R135


olliewalsh (approved):
/lgtm

PR #569, opened 2022-04-22
[OSPK8-430] Introduce additional disk support with v1beta2

stuggi:
1. prepares for new API by split out non version depending parts into a shared API package

2. [OSPK8-430] Introduce additional disk support with v1beta2
    Allows adding additional disks to a vm role, therefore
    restructures disk specification for vmset/ctlplane VM roles.
    Therefore introduces new API verstion for VMSet and Ctlplane.
    Adds OpenStackVMSetDisk type which describes any disk
    configured for the VM.
    Root of vmset now has RootDisk of type OpenStackVMSetDisk
    and AdditionalDisks of type []OpenStackVMSetDisk.
    OpenStackVMSetDisk with the parameters:
    - Name
    - DiskSize
    - StorageClass
    - StorageAccessMode
    - StorageVolumeMode
    - DedicatedIOThread

   Also adds
    * IOThreadsPolicy and BlockMultiQueue to the root vmset spec:
    
    DedicatedIOThread - Disks with dedicatedIOThread set to true will be
    allocated an exclusive thread. This is generally useful if a specific
    Disk is expected to have heavy I/O traffic, e.g. a database spindle.
    
    If IOThreadsPolicy is default, use of IOThreads will be disabled. However,
    if any disk requests a dedicated IOThread, ioThreadsPolicy will be enabled
    and default to shared.
    When ioThreadsPolicy is set to auto IOThreads will also be "isolated" from
    the vCPUs and placed on the same physical CPU as the QEMU emulator thread.
    An ioThreadsPolicy of shared indicates that KubeVirt should use one thread
    that will be shared by all disk devices.
    
    Block Multi-Queue is a framework for the Linux block layer that maps Device
    I/O queries to multiple queues.
    This splits I/O processing up across multiple threads, and therefor multiple
    CPUs. libvirt recommends that the number of queues used should match the
    number of CPUs allocated for optimal performance.
    
    * API update
    
    For existing CRs via the ctlplane mutation webhook parameter from
    old specification gets moved into the new structure. When done
    an annotation gets set on the object to no longer check it.
    
    To auto trigger the storage verstion migration the ctlplane controller
    reconcilation creates the required StorageVersionMigration via
    ensureStorageVersionMigration(). The ctlplane controller then updates/
    patches the owning VMSet objects.
    
    * Validations vi webhook
        On CR create:
            additional disks checks
            - validate Disk Name is not 'rootdisk' as this is reserved for the boot disk
            - validate that disk name is uniq within the VM role
  
       On CR update:
    
            rootdisk checks
            - validate DiskSize don't change
            - validate StorageAccessMode don't change
            - validate StorageClass don't change
            - validate StorageVolumeMode don't change
    
            additional disks checks
            - same as rootdisk update checks
            - validate Disk Name is not 'rootdisk' as this is reserved for the boot disk
            - validate that disk name is uniq within the VM role

3. updates kuttl tests for v1beta2 API changes


api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+		LastTransitionTime: now,
+	}
+	return condition
+}
+
+// Set - Set a particular condition in a given condition list
+func (conditions *ConditionList) Set(conditionType ConditionType, status corev1.ConditionStatus, reason ConditionReason, message string) {
+	condition := conditions.Find(conditionType)
+
+	// If there isn't condition we want to change, add new one
```
bogdando, line 76: can it hold multiple conditions of the same Type? Find would return the first found then.
If it can't, maybe use Type as a dict in the Conditions, instead of a list?

api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+// ConditionDetails used for passing condition information into generic functions
+// e.f. GetDataFromSecret
+type ConditionDetails struct {
+	ConditionNotFoundType   ConditionType
+	ConditionNotFoundReason ConditionReason
+	ConditionErrorType      ConditionType
+	ConditionErrordReason   ConditionReason
+}
+
+// ConditionList - A list of conditions
```
bogdando, line 39: would be nice to explain why it should be a list and cannot be a map with Type as its unique key

api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+	for i, cond := range conditions {
+		if cond.Status == corev1.ConditionTrue {
+			return &conditions[i]
+		}
+	}
+
+	return nil
+}
+
+// UpdateCurrentCondition - update current state condition, and sets previous condition to corev1.ConditionFalse
```
bogdando, line 131: what is the previous one? conditions list doesn't look like a linked one

api/shared/condition_types.go
abays: The Condition list is just the standard pattern for operator status conditions.  It is, however, a reasonable question to wonder why it is a list as opposed to a map.  I noticed that the `CoP` util library you shared with us assumes the same approach for Conditions, but it looks like they may be wrapping some sort of map around the list somehow:
```
  // +patchMergeKey=type
  // +patchStrategy=merge
  // +listType=map
  // +listMapKey=type
  Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```

api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+		condition.Message = message
+		condition.LastTransitionTime = now
+	}
+	condition.LastHeartbeatTime = now
+}
+
+// Find - Check for the existence of a particular condition type in a list of conditions
+func (conditions ConditionList) Find(conditionType ConditionType) *Condition {
+	for i, condition := range conditions {
+		if condition.Type == conditionType {
```
bogdando, line 98: If that would also check for ``status == corev1.ConditionTrue``, I think I would get the point of not using a map, but a dict...

api/shared/condition_types.go
bogdando: I think I got it. Set() finds the "previous" condition of the same type. So the head is the current and the previous one. So a map does not fit. tricky.

api/shared/condition_types.go
bogdando: nvm

api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// UpdateCurrentCondition - update current state condition, and sets previous condition to corev1.ConditionFalse
+func (conditions *ConditionList) UpdateCurrentCondition(conditionType ConditionType, reason ConditionReason, message string) {
+	//
+	// get current condition and update to corev1.ConditionFalse
+	//
+	currentCondition := conditions.GetCurrentCondition()
+	if currentCondition != nil {
+		conditions.Set(
```
bogdando, line 138: so I suppose this should keep the historical trail of a given condition type without getting the list longer. So each condition type would only have at most two instances of it in the list: transition src and dst.

api/shared/condition_types.go
```diff
@@ -0,0 +1,155 @@
...
+			ConditionType(currentCondition.Type),
+			corev1.ConditionFalse,
+			currentCondition.Reason,
+			currentCondition.Message,
+		)
+	}
+
+	//
+	// set new corev1.ConditionTrue condition
```
bogdando, line 147: I think this should be in the else section, or it always overwrites either the only head or the previous condition of it

bogdando: hm, nope, if in the else section, new reason and message would not be written into the current active condition, just leaving it with the same data, and set as inactive... sorry, I don't get all that dance

api/shared/condition_types.go
stuggi: as Andrew said, that's the pattern. the latest condition is the one which is also shown in the web UI, Maybe thats where the list requirement comes from, but I have not checked.

api/shared/condition_types.go
stuggi: this is to set the current latest condition to ConditionFalse, if it exists

api/shared/condition_types.go
stuggi: and then add the new latest condition

api/shared/condition_types.go
stuggi: yes, it just sets/adds/updates the latest condition in the condition list

stuggi:
we can do that, but I'd like to do it in a follow up instead of this PR

api/shared/condition_types.go
stuggi: yes, thats the expected behavior.

api/v1beta1/openstackvmset_webhook.go
```diff
@@ -45,33 +46,33 @@ func (r *OpenStackVMSet) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// +kubebuilder:webhook:verbs=create;update;delete,path=/validate-osp-director-openstack-org-v1beta1-openstackvmset,mutating=false,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackvmsets,versions=v1beta1,name=vopenstackvmset.kb.io,admissionReviewVersions=v1
+//// +kubebuilder:webhook:verbs=create;update;delete,path=/validate-osp-director-openstack-org-v1beta1-openstackvmset,mutating=false,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackvmsets,versions=v1beta1,name=vopenstackvmset.kb.io,admissionReviewVersions=v1
```
abays, line 49: Should `name` here be `vopenstackvmsetv1beta1.kb.io` instead to differentiate from `v1beta2` VMSet webhook?

api/v1beta1/openstackvmset_webhook.go
```diff
@@ -96,10 +97,10 @@ func (r *OpenStackVMSet) ValidateUpdate(old runtime.Object) error {
 func (r *OpenStackVMSet) ValidateDelete() error {
 	vmsetlog.Info("validate delete", "name", r.Name)
 
-	return checkBackupOperationBlocksAction(r.Namespace, APIActionDelete)
+	return CheckBackupOperationBlocksAction(r.Namespace, shared.APIActionDelete)
 }
 
-//+kubebuilder:webhook:path=/mutate-osp-director-openstack-org-v1beta1-openstackvmset,mutating=true,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackvmsets,verbs=create;update,versions=v1beta1,name=mopenstackvmset.kb.io,admissionReviewVersions=v1
+////+kubebuilder:webhook:path=/mutate-osp-director-openstack-org-v1beta1-openstackvmset,mutating=true,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackvmsets,verbs=create;update,versions=v1beta1,name=mopenstackvmset.kb.io,admissionReviewVersions=v1
```
abays, line 103: Same question about name differentiation here

api/v1beta2/openstackcontrolplane_webhook.go
```diff
@@ -0,0 +1,325 @@
...
+	if webhookClient == nil {
+		webhookClient = mgr.GetClient()
+	}
+
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/mutate-osp-director-openstack-org-v1beta2-openstackcontrolplane,mutating=true,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackcontrolplanes,verbs=create;update,versions=v1beta1;v1beta2,name=mopenstackcontrolplane.kb.io,admissionReviewVersions=v1
```
abays, line 59: Similar question regarding `name` here in relation to verions

abays: Ah, but I did see that you had qualified the names of the old version's webhooks, so I guess leaving these as-is won't cause the webhooks to overwrite each other

api/v1beta2/openstackcontrolplane_webhook.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *OpenStackControlPlane) ValidateUpdate(old runtime.Object) error {
+	controlplanelog.Info("validate update", "name", r.Name)
+
+	// Get the OpenStackVMSet object
```
abays, line 266: nit `s/VMSet/ControlPlane`

api/v1beta2/openstackvmset_webhook.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		if len(r.Spec.BootstrapDNS) == 0 && len(osNetCfg.Spec.DNSServers) > 0 {
+			r.Spec.BootstrapDNS = osNetCfg.Spec.DNSServers
+			vmsetlog.Info(fmt.Sprintf("Using BootstrapDNS from %s %s: %v", osNetCfg.GetObjectKind().GroupVersionKind().Kind, osNetCfg.Name, r.Spec.BootstrapDNS))
+		}
+	}
+
+}
+
+//+kubebuilder:webhook:path=/validate-osp-director-openstack-org-v1beta2-openstackvmset,mutating=false,failurePolicy=fail,sideEffects=None,groups=osp-director.openstack.org,resources=openstackvmsets,verbs=create;update;delete,versions=v1beta1;v1beta2,name=vopenstackvmset.kb.io,admissionReviewVersions=v1
```
abays, line 118: Name thing here too

config/webhook/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 resources:
 - manifests.yaml
-#- service.yaml
+- service.yaml
```
abays, line 3: We might want to comment this again, as we don't actually use the webhook service

pkg/openstackbackup/funcs.go
```diff
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	ospdirectorv1beta1 "github.com/openstack-k8s-operators/osp-director-operator/api/v1beta1"
+	ospdirectorv1beta2 "github.com/openstack-k8s-operators/osp-director-operator/api/v1beta1"
```
abays, line 12: s/v1beta1/v1beta2

api/v1beta1/common_openstackcontrolplane.go
```diff
@@ -44,3 +46,44 @@ func GetControlPlane(
...
+			)
+			if err != nil {
+				if k8s_errors.IsNotFound(err) {
+					return uniqNetworksList, fmt.Errorf(fmt.Sprintf("OpenStackNet with NameLower %s not found!", netNameLower))
+				}
+				// Error reading the object - requeue the request.
+				return uniqNetworksList, fmt.Errorf(fmt.Sprintf("Error getting OSNet with labelSelector %v", labelSelector))
+			}
+
+			if _, value := networkList[netNameLower]; !value && network.Spec.VIP {
```
bogdando, line 81: is this supposed to be ``if ok, value := ...; !ok && ...`` ?

bogdando: um, sorry, never-mind :)

abays (approved):
/lgtm

PR #571, opened 2022-04-26
[OSPK8-545] v1.2.x backport - Do not update BMH spec after provisioning

stuggi:
Update of BMH objects of provisioned nodes can result in
reprovisioning of the nodes. This change only updates the
BMH spec only on initial deployment when there is no consumer
ref set.
Updating labels is done on any change, which might be needed
on future operator changes.

Also updates caused by controller-gen 0.8.0 and go.mod updates


abays (approved):
/lgtm

abays (approved):
/lgtm

PR #574, opened 2022-04-29
Fix openstackclient restart loop

stuggi:
There can be a pod re-create loop in case the image of a
openstackclient changes, but the pod status does not reflect
the change of the new image in time before the new
reconcilation.

This changes to update the openstackclient image only if the
spec in the image changes and it is a pod update.


olliewalsh (approved):
/lgtm

PR #577, opened 2022-05-03
[OSPK8-467 OSPK8-472 OSPK8-474] enhance validation for osnetcfg

stuggi:
Adds additional validation to check the osnetcfg:

* verifies the a network is tagged as the ctlplane network
* verifies information provided in ipv4/ipv6 section are
  actual ipv4/ipv6 information
* verifies that information in the ipv4 section is pure ipv4
* verifies ips provided for AllocationStart, AllocationEnd and
  Gateway are part of the network set via the Cidr
* verifies that in a subnet definition there is not both ipv4
  and ipv6 provided

### DIFF
--- a/api/shared/net.go
+++ b/api/shared/net.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2020 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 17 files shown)


abays (approved):
/lgtm

PR #579, opened 2022-05-04
Fix compute/osbms hostname to be FQDN

stuggi:
DomainName was not correct to render the osbms userdata. Therefore
the DomainName was always empty and not set, which resulted in
osbms nodes to default to short hostname:

```
[root@computehci-1 ~]# hostname
computehci-1
[root@computehci-1 ~]# hostname --short
computehci-1
[root@computehci-1 ~]# hostname --long
computehci-1.osptest.test.metalkube.org
```

And osvmset nodes to fqdn:
```
[root@controller-0 ~]# hostname
controller-0.osptest.test.metalkube.org
[root@controller-0 ~]# hostname --short
controller-0
[root@controller-0 ~]# hostname --long
controller-0.osptest.test.metalkube.org
```

This sets the DomainName correct for the templateParameters that
the osbms cloud-init userdata has the DomainName information set.


abays (approved):
/lgtm

PR #581, opened 2022-05-09
[v1.2.x backport] Bump containerd to avoid CVE detection with earlier versions

stuggi:
while it is just imported as dependency, bump containerd version to avoid CVE [1] detection

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2011007

(cherry picked from commit 641e3fc9ef5972fd1602fcaba53ccf730c746eb4)


abays (approved):
/lgtm

PR #583, opened 2022-05-11
[OSPK8-398] anti-affinity placement of controller VMs

stuggi:
Set default affinity rules that two pods of the same role should
not run if possible on the same worker node. With this it is still
possible to live migrate VMs and they run on the same node, but
on the next scheduling request (live migration) they get distributed
again.

### DIFF
--- a/pkg/common/ipam.go
+++ b/pkg/common/ipam.go
@@ -203,9 +203,7 @@ func BigIntToIP(inipint big.Int) net.IP {
 	intbytes := inipint.Bytes()
 	if len(intbytes) == net.IPv6len {
 		// This is an IPv6 address.
-		for i := 0; i < len(intbytes); i++ {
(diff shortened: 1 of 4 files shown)


dprince:
@stuggi I think its okay. The one question I would have is would we ever want the ability to control this more specifically with more generic labels. Like say I have 3 workers in my (larger) OCP cluster and I want the 3 VMs there. We'd still need a config to be able to accommodate this.

stuggi:
> Set default affinity rules that two pods of the same role should
> not run if possible on the same worker node. With this it is still
> possible to live migrate VMs and they run on the same node, but
> on the next scheduling request (live migration) they get distributed
> again.

This is already possible. you can specify a nodeselector in the osnetcfg for the nncp config to configure network devices on the worker nodes. The VMs get scheduled to those nodes.

PR #584, opened 2022-05-12
NodeSelector for VMSet placement

stuggi:
Addes NodeSelector parameter to VMSet and Ctlplane. With this it
is possible to control the target worker nodes for VM placement
on a per role base.


abays (commented):
+1 from me, but deferring to @dprince

dprince (approved):

PR #585, opened 2022-05-16
Set BootOrder: 1 on rootdisk

stuggi:
To prevent an issue when the rootdisk is not the first
disk in the disk list, set BootOrder: 1 [1] on the rootdisk
as this is always the disk the VM should boot from.

[1] https://pkg.go.dev/kubevirt.io/api@v0.51.0/core/v1#Disk


abays (approved):
/lgtm

PR #586, opened 2022-05-23
Add CtlplaneNetworkAttribues for 16.2

olliewalsh:
Tripleoclient now sets this in stable/train - https://review.opendev.org/c/openstack/python-tripleoclient/+/825689

### DIFF
--- a/templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
+++ b/templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
@@ -124,3 +124,36 @@ parameter_defaults:
 {{- end }}
 {{- end }}
 {{- end }}
+  #
+  # CtlplaneNetworkAttributes
+  CtlplaneNetworkAttributes:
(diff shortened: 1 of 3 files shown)


templates/openstackconfiggenerator/config/16.2/rendered-tripleo-config.yaml
```diff
@@ -123,4 +123,38 @@ parameter_defaults:
...
 {{- end }}
 {{- end }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+  #
+  # CtlplaneNetworkAttributes
+  CtlplaneNetworkAttributes:
+    {{- range $netname, $net := .NetworksMap }}
+    {{- $netname_lower := lower $net.Name }}
```
stuggi, line 131: netname_lower is not used and should be not required as there is $net.NameLower? Think we can also remove it from the 17.0 templates

olliewalsh: done

stuggi (approved):
/lgtm

PR #587, opened 2022-05-23
Add default services to 17.0 or wallaby

stuggi:
Right now the webhook only adds the default service list to
OSPVersion 17.0, but also need to be added for upstream wallaby.


olliewalsh (approved):
/lgtm

PR #588, opened 2022-06-15
Bump ignition/v2

dprince:


stuggi:
for reference, golangci issue fixed in https://github.com/openstack-k8s-operators/osp-director-operator/pull/589

PR #590, opened 2022-06-23
Fix default route handling in VM nic template

olliewalsh:
Change the VM nic config template to a tripleo role template as it
needs to get the default_route_networks value from roles_data.yaml.

Update the template to only set default routes on the specified
interfaces.

### DIFF
--- a/templates/openstackconfiggenerator/config/16.2/nic/net-config-multi-nic.role.j2.yaml
+++ b/templates/openstackconfiggenerator/config/16.2/nic/net-config-multi-nic.role.j2.yaml
@@ -119,9 +119,11 @@ resources:
                 routes:
                   list_concat_unique:
                     - get_param: ControlPlaneStaticRoutes
+{{`{%- if role.default_route_networks is not defined or 'ControlPlane' in role.default_route_networks %}`}}
                     - - default: true
                         next_hop:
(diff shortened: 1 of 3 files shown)


templates/openstackconfiggenerator/config/16.2/nic/net-config-multi-nic.role.j2.yaml
```diff
@@ -138,6 +140,11 @@ resources:
                   list_concat_unique:
                     - get_param: {{ $net.Name }}Routes
                     - get_param: {{ $net.Name }}InterfaceRoutes
+{{`{%- if '`}}{{ $net.Name }}{{`' in role.default_route_networks %}`}}
```
stuggi, line 143: should we also have/keep the condition from [1] in L51 for the {{ $net.Name }}InterfaceDefaultRoute ?

[1] https://github.com/openstack/tripleo-heat-templates/blob/stable/train/network/config/multiple-nics/role.role.j2.yaml#L62

olliewalsh: Doesn't really matter since the param has a default value

stuggi (approved):
/lgtm

PR #593, opened 2022-07-01
Require baremetal-operator/apis (security fix)

dprince:
This swaps out the baremetal-operator requirement for
baremetal-operator/apis (also a go module). This
avoids the yujunz/go-getter requirement in go.sum
and thus avoids security warnings against this
unpatched package.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2101027


abays (approved):
/lgtm

PR #596, opened 2022-07-04
ServiceVIP to be optional in IP reservation

stuggi:
ServiceVIP is right now a required parameter in the IPReservation
type. This prevents update the operator in existing environments
from v1.2.x to later version as there is a validation if existing
object meet the requirements.
ServiceVIP is only used in OSP17 [1] environments and therefore
can be set to optional.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/controllers/openstackcontrolplane_controller.go#L885


abays (approved):
/lgtm

PR #598, opened 2022-07-06
Fix VM role nic template

stuggi:
Since the switch to use role template for the virtual machines, network
configuration is broken when using multiple VM roles which don't use
the same networks. Basically at the moment the roles nic template still
renders the networks for the roles which results
nic-template.role.j2.yaml has fixed network parameter list of the last
vm role instead of the jinja templating. Therefore process-templates.py
does not create the correct nic template for the specified networks
for the role.

This changes the net-config-multi-nic.role.j2.yaml to be almost raw
ooo roles nic template. The only go templated is to set the
ClusterServiceIP for the pod interface to be able to connect to the
OCP cluster for fencing.

Note: templated information needs to be translated from
{{network.name_lower}} to using golang raw string constant like
{{`{{network.name_lower}}`}}.


templates/openstackconfiggenerator/config/16.2/nic/net-config-multi-nic.role.j2.yaml
```diff
@@ -119,68 +131,71 @@ resources:
...
                     - - default: true
                         next_hop:
                           get_param: ControlPlaneDefaultRoute
-{{`{%- endif %}`}}
-{{- $nic_counter := 2 }}
-{{- range $netid, $net := .Role.Networks }}
-{{- if and (ne $net.Name "External") (ne $net.Name "Tenant") (ne $net.NameLower "ctlplane") }}
-{{- $nic_counter = add $nic_counter 1 }}
+{%- endif %}
+{%- set nics_used = [2] %}
```
abays, line 139: This is used at the end of the loop, but I'm sure why it is needed when we use `loop.index + 2` (seemingly) instead.  Could you explain what is happening here?

stuggi: ah we can remove that. It comes from the default multi nic role template[1] where ad the end they add nics for the ComputeOvsDpdk role[2]. In our case the `loop.index + 2` should be enough. I'll update it.

[1] https://github.com/openstack/tripleo-heat-templates/blob/stable/train/network/config/multiple-nics-vlans/role.role.j2.yaml#L137
[2]https://github.com/openstack/tripleo-heat-templates/blob/stable/train/network/config/multiple-nics-vlans/role.role.j2.yaml#L206

abays (approved):
/lgtm

olliewalsh (approved):

PR #603, opened 2022-07-15
[OSPK8-601] fix ansible-inventory jq query

stuggi:
The current jq query does not filter deleted hosts correct. Instead
it checks if any node is not deleted and then adds all of the nodes
in a role. This changes the query to delete the deleted marked nodes
from the result.


stuggi:
https://github.com/openstack-k8s-operators/osp-director-operator/pull/605 should fix the continuous-integration/jenkins/pr-merge issue

PR #604, opened 2022-07-15
[OSPK8-600] Don't delete role reservations with PreserveReservations true

stuggi:
In master of the osp-director-operator OSIPSet got introduced.
During the update from 1.2.x to latest the OSIPSets get created.
The OSNetConfig depends on the OSIPSet to identify which
reservations could be deleted. If an OSIPSet is gone the OSNetConfig
expects that the reservations for the whole role could be delted.

This conflicts during the update as there is no OSIPSet and therefore
the OSNetConfig will delete all reservations, only the nodes which
have static reservations are save to keep their current IPs.

This changes the OSNetConfig behaviour to not delete reservations
for deleted roles if Spec.PreserveReservations is set to true,
which is the default behaviour.

### DIFF
--- a/controllers/openstacknetconfig_controller.go
+++ b/controllers/openstacknetconfig_controller.go
@@ -1250,10 +1250,12 @@ func (r *OpenStackNetConfigReconciler) ensureIPReservation(
 	}
 
 	//
-	// cleanup reservations from deleted roles
+	// cleanup reservations from osnet RoleReservations if
+	// * role is not in allRoles (ipset deleted)
+	// * and PreserveReservations is not set
 	//
 	for role := range osNet.Spec.RoleReservations {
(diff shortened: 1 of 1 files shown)


stuggi:
I first had a different fix [1], which checks in the osipset if it is a new ipset to check the owning resource based on the owner reference and get the host reservations from there as a one time action, but I think this is much simpler and ok to not delete the reservations for gone roles if preservereservation is set.

[1] https://github.com/stuggi/osp-director-operator/tree/fix_ipset_create

abays (approved):
/lgtm

stuggi:
https://github.com/openstack-k8s-operators/osp-director-operator/pull/605 should fix the continuous-integration/jenkins/pr-merge issue

PR #611, opened 2022-08-03
allign heat tunings with traditional undercloud

stuggi:
Current config version made playbook create to fail on slower/busy ocp cluster.
With these changes taken from a 16.2 traditional undercloud it worked fine.

### DIFF
--- a/templates/openstackephemeralheat/bin/heat.conf
+++ b/templates/openstackephemeralheat/bin/heat.conf
@@ -9,6 +9,11 @@ deferred_auth_method=password
 max_nested_stack_depth=10
 max_json_body_size=8388608
 default_deployment_signal_transport=HEAT_SIGNAL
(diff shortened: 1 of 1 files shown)


templates/openstackephemeralheat/bin/heat.conf
```diff
@@ -9,6 +9,11 @@ deferred_auth_method=password
 max_nested_stack_depth=10
 max_json_body_size=8388608
 default_deployment_signal_transport=HEAT_SIGNAL
+max_resources_per_stack=-1
+convergence_engine=True
+max_nested_stack_depth=7
+num_engine_workers=4
```
dprince, line 15: In addition to this we also have the option to scale Heat Engine (not workers) on OpenShift. But this is good to.

dprince (approved):

PR #612, opened 2022-08-08
Fix multi subnet routes ... again

stuggi:
This was missed when we switched to role nic template a while ago.

As mentioned in [1]:
The OSP 16.2 tripleo nic templates have the <netName>InterfaceRoutes parameter
per default included. The routes parameter rendered in
environments/network-environment.yaml which are named <netName>Routes get
usually set on the neutron network host_routes property and get added to the
role <netName>InterfaceRoutes parameter.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/pull/529


abays (approved):
/lgtm

olliewalsh (approved):

PR #616, opened 2022-08-30
KUTTL and backup/restore fixes

abays:
Fixes two KUTTL tests and also fixes backup/restore (backup was working but restore needed tweaking to work with the newer networking CRDs)


controllers/openstackbackuprequest_controller.go
```diff
@@ -363,16 +363,22 @@ func (r *OpenStackBackupRequestReconciler) ensureLoadBackup(
...
 	// Create all CRs in the spec first, and set their status to some initial state
+	// NOTE: For some reason, the "status" of the various network-related CRs is being
+	//       removed after the CR is applied.  To work around this, we create a deep
+	//       copy of the status to apply after the CR's spec has been applied.
 
 	msg := fmt.Sprintf("OpenStackBackup %s initial load", backup.Name)
 
 	// OpenStackNets
 	for _, item := range backup.Spec.Crs.OpenStackNets.Items {
+		status := item.Status.DeepCopy()
```
dprince, line 374: Looking at the code it seems that we use CreateOrPatch to restore each object. I wonder if that just doesn't set the status correctly. Also, had a question if on restoration if we should be calling a controllerutil.SetControllerReference. Perhaps I missed this being done elsewhere though

dprince (approved):

PR #617, opened 2022-09-02
Sdk v1.23.0/golang 1.18 upgrade

dprince:


config/crd/bases/osp-director.openstack.org_openstackbackuprequests.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.9.0
```
abays, line 6: Didn't we run into some sort of issue with `v0.9.0` with Podified such that we immediately bumped to `v0.9.2`?  If so, just wondering if we want to do that here too

dprince:
Pushing this to address the Prow golangci failure: https://github.com/openshift/release/pull/31969

config/crd/bases/osp-director.openstack.org_openstackbackuprequests.yaml
dprince: I'll bump it. Wasn't strictly required for this update but yes we've done it elsewhere and won't harm anything.

dprince: I had a cached version in bin/controller-gen. Removed and re-ran make. All good now I think

Makefile
```diff
@@ -242,7 +245,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.19.1/$${OS}-$${ARCH}-opm ;\
```
olliewalsh, line 245: Did you mean to drop the ;\

dprince: @olliewalsh I did not. Thanks! Should be fixed now.

olliewalsh (approved):
/lgtm

PR #619, opened 2022-09-09
Add dependabot config

dprince:
Initial dependabot config setup to ignore github.com/coreos/ignition
which we don't use directly.


abays (approved):
/lgtm

PR #638, opened 2022-09-14
[OSPK8-626] Fix WaitOnJob when a job is retried

olliewalsh:
Need to check the Succeeded count before checking the Failed count when determining the job status.


stuggi (approved):
/lgtm

PR #642, opened 2022-09-20
Change RunStrategy of VMs to manual

stuggi:
Fencing can fail with current RunStrategy RerunOnFailure because
fencing is not a graceful shutdown and kubevirt will jump in and
restart immediately the VM, but fence_kubevirt is waiting for
the successful shutdown state of the VM befor triggering the start.

This changes the RunStrategy to manual and only triggers a start
of the VM once when the VM CR got created. After that its the
responsibility if the end user, or pacemaker to manage the power
state of the VMs.

Resolves: rhbz#2126727


abays (approved):
/lgtm

PR #643, opened 2022-09-21
Revert to kubevirt v0.53.2 to match OCP4.11 virt

stuggi:
OCP 4.11 virtualization is based on KubeVirt v0.53.2 [1].Lets revert to this version instead of upstream latest.

[1] https://docs.openshift.com/container-platform/4.11/virt/virtual_machines/virt-create-vms.html#additional-resources_virt-create-vms_virt-create-vms


abays:
Just curious as to why 4.11 and not 4.10.  Are we targeting 4.11 for GA now?

stuggi:
> Just curious as to why 4.11 and not 4.10. Are we targeting 4.11 for GA now?

no, but returning to 0.49.0 requires a lot more revert of updated modules on go.mod. We can do that, but there is some go.mod hassle involved :) . As we don't use enhanced features from later versions, I think we could pin to the 4.11 used version. But that's up for discussion here.

abays:
> > Just curious as to why 4.11 and not 4.10. Are we targeting 4.11 for GA now?
> 
> no, but returning to 0.49.0 requires a lot more revert of updated modules on go.mod. We can do that, but there is some go.mod hassle involved :) . As we don't use enhanced features from later versions, I think we could pin to the 4.11 used version. But that's up for discussion here.

This seems fine to me, then

abays (approved):
/lgtm

PR #652, opened 2022-10-17
Move EnableFencing to OpenStackConfigGenerator

stuggi:
The EnableFencing parameter is only used in the ConfigGenerator to auto create templates to enable fencing. Therefore it makes sense to have this parameter in the ConfigGenerator CRD instead of the ControlPlane.

Also as it is a required parameter lets not specify a default and have the user to explicitely either enable or disable it so that the user is being forced to set either state.

### DIFF
--- a/config/crd/bases/osp-director.openstack.org_openstackconfiggenerators.yaml
+++ b/config/crd/bases/osp-director.openstack.org_openstackconfiggenerators.yaml
@@ -46,6 +46,13 @@ spec:
             description: OpenStackConfigGeneratorSpec defines the desired state of
               OpenStackConfigGenerator
             properties:
+              enableFencing:
+                description: 'EnableFencing allows the automatic creation of required
+                  heat environment files to enable fencing. Note: - Production OSP
+                  environments MUST have fencing enabled. - Requires the fence-agents-kubevirt
+                  package to be installed in the virtual machines for the roles running
+                  pacemaker.'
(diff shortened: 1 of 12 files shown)


stuggi:
switched to draft to not merge it. I want to test something before doing that for update

olliewalsh (approved):
/lgtm

config/crd/bases/osp-director.openstack.org_openstackconfiggenerators.yaml
```diff
@@ -46,6 +46,13 @@ spec:
             description: OpenStackConfigGeneratorSpec defines the desired state of
               OpenStackConfigGenerator
             properties:
+              enableFencing:
+                description: 'EnableFencing allows the automatic creation of required
+                  heat environment files to enable fencing. Note: - Production OSP
```
olliewalsh, line 51: should we default to true since it's required for prod?

stuggi: we could, I thought it might also be good that this is a must set parameter either true/false as it also requires to install fence-agents-kubevirt.

olliewalsh: ok, makes sense

olliewalsh (approved):
/lgtm

PR #653, opened 2022-10-19
[v1.3.x backport] Move EnableFencing to OpenStackConfigGenerator

stuggi:
The EnableFencing parameter is only used in the ConfigGenerator to auto create templates to enable fencing. Therefore it makes sense to have this parameter in the ConfigGenerator CRD instead of the ControlPlane.

Also as it is a required parameter lets not specify a default and have the user to explicitely either enable or disable it so that the user is being forced to set either state.

(cherry picked from commit 6dc6f8b04ba6b30e021893947803b5b666e8d0fb)

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,6 +613,7 @@ Create a base RHEL data volume prior to deploying OpenStack.  This will be used 
       name: default
       namespace: openstack
     spec:
+      enableFencing: False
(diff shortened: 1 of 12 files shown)


abays (approved):
/lgtm

PR #654, opened 2022-10-20
Allow updating password secret with new/missing passwords

stuggi:
Right now the tripleo password secret only gets created once if it does not exist. If a new operator version introduces a new password, this would not get added to the password secret of an existing environment.

This change validates if the password list is missing any entry on the top level. If there is one missing it merges the new password map[string]interface{} into the existing. It only validates if the password list is complete, using the highest level password keys.

### DIFF
--- a/controllers/openstackcontrolplane_controller.go
+++ b/controllers/openstackcontrolplane_controller.go
@@ -18,11 +18,14 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
+	"gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


stuggi:
moving to draft to finish some more testing

controllers/openstackcontrolplane_controller.go
```diff
@@ -478,13 +463,135 @@ func (r *OpenStackControlPlaneReconciler) createOrGetTripleoPasswords(
...
+			return err
+		}
+
+		// recursice convert nested map[interface{}]interface{} password data
+		// into map[string]interface{} and merge into a new tripleoPasswords map
+		tripleoPasswords, err := common.RecursiveMergeMaps(
+			map[string]interface{}{},
+			tripleoPasswordsRaw,
+			common.PasswordMaxDepth,
+		)
```
abays, line 485: Just want to make sure I understand the comment.  So the `tripleoPasswordsRaw` variable is `map[string]interface{}` but the `interface{}` values themselves are `map[interface{}]interface{}`, so it's really `map[string]map[interface{}]interface{}?  That's my guess, at least, because it seems like you already have `map[string]interface{}`.

stuggi: yes correct, nested maps are of type map[interface{}]interface{}, but how many nested maps there are is unknown.

controllers/openstackcontrolplane_controller.go
```diff
@@ -478,13 +463,135 @@ func (r *OpenStackControlPlaneReconciler) createOrGetTripleoPasswords(
...
+					return err
+				}
+
+				currentPasswordsMap[key] = string(b)
+			}
+		}
+
+		// validate if new password entries need to be added
+		newPasswordsMap := make(map[string]interface{})
+		for key := range common.GeneratePasswordsMap() {
```
olliewalsh, line 526: why not iterate passwordNames?

stuggi: right, will make it public so that it can be used here.

controllers/openstackcontrolplane_controller.go
```diff
@@ -478,13 +463,135 @@ func (r *OpenStackControlPlaneReconciler) createOrGetTripleoPasswords(
...
+		if err := yaml.Unmarshal(tripleoPasswordsByte, &tripleoPasswordsRaw); err != nil {
+			cond.Message = fmt.Sprintf("Error extract TripleoPasswords from Secret %s", controlplane.TripleoPasswordSecret)
+			cond.Reason = shared.ControlPlaneReasonTripleoPasswordsSecretError
+			cond.Type = shared.CommonCondTypeError
+			err = common.WrapErrorForObject(cond.Message, instance, err)
+
+			return err
+		}
+
+		// recursice convert nested map[interface{}]interface{} password data
```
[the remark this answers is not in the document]
stuggi: done

controllers/openstackcontrolplane_controller.go
```diff
@@ -478,13 +463,135 @@ func (r *OpenStackControlPlaneReconciler) createOrGetTripleoPasswords(
...
+					cond.Message = fmt.Sprintf("Error marshal nested password maps from TripleoPasswords from Secret %s", controlplane.TripleoPasswordSecret)
+					cond.Reason = shared.ControlPlaneReasonTripleoPasswordsSecretError
+					cond.Type = shared.CommonCondTypeError
+					err = common.WrapErrorForObject(cond.Message, instance, err)
+
+					return err
+				}
+
+				currentPasswordsMap[key] = string(b)
+			}
```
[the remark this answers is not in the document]
olliewalsh: Lets hope a password param is not added that expects a list :-)

olliewalsh (approved):

controllers/openstackcontrolplane_controller.go
dprince: I would have thought map[string]interface{} would be enough here. Didn't realize we were nesting that deep

stuggi: its the nested jsons which gets marshalled into nested map[interface{}]interface{}.

PR #655, opened 2022-10-20
Align kube-rbac-proxy versions

dprince:
This updates the relatedImages version to match the one we use in config/default/manager_auth_proxy_patch.yaml.

TODO: look into adding a common ENV variable for these


stuggi (approved):
/lgtm also needs a v1.3.x backport

abays (approved):
/lgtm

PR #656, opened 2022-10-21
Truncate configVersion diff to < 2M

dprince:


pkg/openstackconfigversion/git_util.go
```diff
@@ -228,3 +228,12 @@ func SyncGit(
 
 	return configVersions, nil
 }
+
+// truncateDiff  truncate the diff size to less than 2MB
+func truncateDiff(diff string, log logr.Logger) string {
+	if len(diff) > 2097151 {
```
stuggi, line 234: two things:
* the max message size is 2MB, so we need to be lower than that for other object data fields
* while the max massage send size is 2MB, the default max size we can store in etcd is 1MB. maybe we should just truncate to 512KB.

pkg/openstackconfigversion/git_util.go
```diff
@@ -228,3 +228,12 @@ func SyncGit(
...
+		log.Info("Git diff was truncatated as it is larger than 2MB.")
+		return diff[:2097151]
```
stuggi, line 236: I'd also add that info to the object diff data so something like:

`return fmt.Sprintf("Git diff was truncated to 512KB.\n %s", diff[:524800])`

stuggi (approved):
/lgtm

PR #660, opened 2022-10-24
Fix IsReady cond reason

stuggi:
Creating a backup is stuck in quiescing because IsReady() from OSVMSet is using OpenStackVMSetProvisioned instead of VMSetCondReasonVirtualMachineProvisioned as set in the controller at
https://github.com/openstack-k8s-operators/osp-director-operator/blob/v1.3.x/controllers/openstackvmset_controller.go#L1703

This alligns IsReady to use VMSetCondReasonVirtualMachineProvisioned.


abays (approved):
/lgtm

olliewalsh (approved):

PR #671, opened 2022-11-02
Filter out unused t-h-t roles when generating config

olliewalsh:
Container image prepare skips any roles with Count=0 but overcloud.j2.yaml does not, resulting in heat errors due to missing image params.

Removing the unused roles from roles_data.yaml avoids the issue and should make the heat stack create more efficient.

### DIFF
--- a/templates/openstackconfiggenerator/bin/create-playbooks.sh
+++ b/templates/openstackconfiggenerator/bin/create-playbooks.sh
@@ -57,8 +57,12 @@ sed -e "s|/usr/share/openstack\-tripleo\-heat\-templates|\.|" -i $HOME/config-tm
 cp -a $HOME/config-tmp/* "$TEMPLATES_DIR/"
 
 pushd $TEMPLATES_DIR
-python3 tools/process-templates.py -r $TEMPLATES_DIR/roles_data.yaml -n $TEMPLATES_DIR/network_data.yaml
 
+# Remove unused roles from roles_data.yaml as container image prepate skips roles with count=0 but
+# process-templates.py currently does not
+/home/cloud-admin/process-roles.py -r $TEMPLATES_DIR/roles_data.yaml -e rendered-tripleo-config.yaml
+
+python3 tools/process-templates.py -r $TEMPLATES_DIR/roles_data.yaml -n $TEMPLATES_DIR/network_data.yaml
 
 # NOTE: only applies to OSP 16, on OSP 17+ we set NetworkSafeDefaults: false in the Heat ENV
 if [ -e ./network/scripts/run-os-net-config.sh ]; then
--- a/templates/openstackconfiggenerator/bin/process-roles.py
+++ b/templates/openstackconfiggenerator/bin/process-roles.py
@@ -1,0 +1,167 @@
+#! /usr/bin/env python3
+
+import argparse
+import collections
+import logging
+import sys
+import unittest
+import unittest.mock
+import yaml
+
+logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+LOG = logging.getLogger('process-roles')
+
+
+class TemplateLoader(yaml.SafeLoader):
+    def construct_mapping(self, node):
+        self.flatten_mapping(node)
+        return collections.OrderedDict(self.construct_pairs(node))
+
+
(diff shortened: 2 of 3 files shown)


templates/openstackconfiggenerator/bin/process-roles.py
```diff
@@ -0,0 +1,77 @@
...
+
+def filter_roles(roles_file, environment):
+    role_counts = get_role_counts(environment)
+
+    with open(roles_file, 'r', encoding='utf-8') as orig:
+        original_data = orig.read()
+        original_roles = yaml.load(original_data, Loader=TemplateLoader)
+
+    # Omit roles that are not in use or count is 0
+    new_roles = [x for x in original_roles if role_counts.get(x['name'], 0) > 0]
```
stuggi, line 53: for debugging purposes, should we add here some logging output that role bla got removed because it is not in use or count ==0?

templates/openstackconfiggenerator/bin/create-playbooks.sh
```diff
@@ -57,8 +57,12 @@ sed -e "s|/usr/share/openstack\-tripleo\-heat\-templates|\.|" -i $HOME/config-tm
 cp -a $HOME/config-tmp/* "$TEMPLATES_DIR/"
 
 pushd $TEMPLATES_DIR
-python3 tools/process-templates.py -r $TEMPLATES_DIR/roles_data.yaml -n $TEMPLATES_DIR/network_data.yaml
 
+# Remove unused roles from roles_data.yaml as container image prepate skips roles with count=0 but
```
stuggi, line 61: this is really good, as it also fixes the situation where you copy a default role, which might not be used, but has a default count >0. That results in an error as you don't have a bmset or vmset for it.

abays (commented):
+1

templates/openstackconfiggenerator/bin/process-roles.py
olliewalsh: Added logging and basic unittest

templates/openstackconfiggenerator/bin/process-roles.py
```diff
@@ -0,0 +1,167 @@
...
+def get_role_counts(environment):
+    """Find role counts in heat environment files"""
+    role_counts = {}
+    for env_file in environment:
+        try:
+            with open(env_file, 'r', encoding='utf-8') as envfile:
+                data = yaml.safe_load(envfile.read())
+            if data is not None:
+                for param, val in data.get('parameter_defaults', {}).items():
+                    if param.endswith('Count'):
```
stuggi, line 48: from a quick grep there are some parameters (DockerPuppetProcessCount, CinderXtremioArrayBusyRetryCount, IronicDhcpv6StatefulAddressCount, KeystoneUniqueLastPasswordCount, CollectdSensubilityWorkerCount) which also end with Count. There is a chance we get those added to role_counts. we could first get all role names from roles_file and pass them here.

templates/openstackconfiggenerator/bin/process-roles.py
```diff
@@ -0,0 +1,167 @@
...
+    role_counts = get_role_counts(environment)
+
+    LOG.info("Loading original role data from '%s'", roles_file)
+    with open(roles_file, 'r', encoding='utf-8') as orig:
+        original_data = orig.read()
+        original_roles = yaml.load(original_data, Loader=TemplateLoader)
+
+    # Omit roles that are not in use or count is 0
+    new_roles = []
+    for role in original_roles:
```
stuggi, line 80: The above mentioned situation should get filtered out here.

stuggi:
@olliewalsh do you think we need to handle above mentioned situation? I think it should not impact the result

stuggi:
@olliewalsh do you think we need to handle above mentioned situation? I think it should not impact the result

templates/openstackconfiggenerator/bin/process-roles.py
olliewalsh: Yes, false matches were a concern. If it's not an integer then we ignore it completely. Keeping these params that look like a role count should not be an issue since we only lookup the counts for the roles defined in roles_data

stuggi (approved):
/lgtm

PR #678, opened 2022-11-17
[OSPL8-641/OSPK8-642] Add script to cleanup stale baremetalset compute entries

stuggi:
If the count of an OSBms gets increased above the number of available BMH the OSBms will have compute nodes entries stuck with `hostRef: unassigned`, even when reduce the spec.count again to a number that is available.

This adds a script which can be used to cleanup compute entries from status of osbmset, osnet, osipset and osnetconfig. Also the reservations from the spec of osnet.

### DIFF
--- a/scripts/baremetalset_cleanup_stale_entries.sh
+++ b/scripts/baremetalset_cleanup_stale_entries.sh
@@ -1,0 +1,115 @@
+#!/bin/bash
+# USAGE: ./script.sh <ROLE name - case sensitive> <list of compute names to remove>
+#        e.g. ./script.sh Compute compute-2 compute-3
+
+ROLE=$1
+COMPUTES="${@:2}"
+
+if [ -z "$ROLE" ]; then
+  echo "Please pass the role name as first argument"; exit 1
+fi
+
+if [ -z "$COMPUTES" ]; then
+      echo "Please pass list of compute names to remove"; exit 1
+fi
+
+#
+# Disable the operator while updating the status
+#
+CSV_NAME=$(oc get csv -n openstack  --output=jsonpath={.items..metadata.name})
+oc patch -n openstack csv/$CSV_NAME --type='json' \
+--patch='[{"op":"replace", "path":"/spec/install/spec/deployments/0/spec/replicas", "value":0}]'
+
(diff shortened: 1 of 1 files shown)


scripts/baremetalset_cleanup_stale_entries.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+# USAGE: ./script.sh <ROLE name - case sensitive> <list of compute names to remove>
+#        e.g. ./script.sh Compute compute-2 compute-3
+
```
olliewalsh, line 4: Is it worth `set -eu` and `-o pipefail` to fail fast e.g if curl/jq are not in the $PATH

scripts/baremetalset_cleanup_stale_entries.sh
```diff
@@ -0,0 +1,115 @@
...
+ROLE=$1
+COMPUTES="${@:2}"
+
+if [ -z "$ROLE" ]; then
+  echo "Please pass the role name as first argument"; exit 1
```
olliewalsh, line 9: I think it would be more helpful to print the full USAGE string

scripts/baremetalset_cleanup_stale_entries.sh
```diff
@@ -0,0 +1,115 @@
...
+fi
+
+if [ -z "$COMPUTES" ]; then
+      echo "Please pass list of compute names to remove"; exit 1
```
olliewalsh, line 13: I think it would be more helpful to print the full USAGE string

scripts/baremetalset_cleanup_stale_entries.sh
```diff
@@ -0,0 +1,115 @@
...
+      echo patching osnet $OSNET spec index $INDEX
+      oc patch osnet $OSNET --type=json -p="[{\"op\": \"remove\", \"path\": \"/spec/roleReservations/$ROLE/reservations/$INDEX\"}]" -n openstack
+    fi
+  done
+done
+
+#
+# Stop the kube proxy
+#
+kill $PROXY_PID
```
olliewalsh, line 115: Use a trap to always run this?

dprince (approved):

PR #684, opened 2022-11-17
[OSPK8-641] Add webhook logic to validate OSBMS count field

abays:


api/shared/labels.go
```diff
@@ -32,4 +32,10 @@ const (
 	// OpenStackNetConfigReconcileLabel - label set on objects on which change
 	// trigger a reconcile of the osnetconfig
 	OpenStackNetConfigReconcileLabel = "osnetconfig-ref"
+	// OpenStackBaremetalSetAppLabel -
+	OpenStackBaremetalSetAppLabel = "osp-baremetalset"
+	// OwnerControllerNameLabelSelector -
+	OwnerControllerNameLabelSelector string = "osp-director.openstack.org/controller"
+	// OwnerUIDLabelSelector -
+	OwnerUIDLabelSelector string = "osp-director.openstack.org/uid"
```
abays, line 40: I know this is duplicating what we have in the main `common` package.  Moving all the constants to the API just seems like a lot of work for this initial pass.

stuggi (commented):
looks good to me. I think that covers what we discussed. maybe add [OSPK8-641] as prefix for the jira we track it.

stuggi (approved):
/lgtm

PR #688, opened 2022-11-22
Prevent OSBMS reconciliation when in-use BMHs have disappeared

abays:
Prevents processing of scale-up/scale-down for a given OSBMS instance if one or more BMHs in its `status.baremetalHosts`'s `hostRef`s has been manually (inappropriately) deleted.


api/v1beta1/common_openstackbaremetalset.go
```diff
@@ -70,6 +70,48 @@ func getDeletionAnnotatedBmhHosts(
...
+		ctx,
+		c,
+		"openshift-machine-api",
+		map[string]string{},
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, bmhStatus := range instance.Status.DeepCopy().BaremetalHosts {
```
stuggi, line 93: it won't hurt, just wondering if we need to use here DeepCopy()?

abays:
Added KUTTL tests as well

tests/kuttl/common/scripts/prep.sh
```diff
@@ -12,21 +12,27 @@ oc delete openstackclient --all -n openstack
 oc delete openstackbaremetalset --all -n openstack
 oc delete openstackprovisionserver --all -n openstack
 
+# Make sure BMHs exist (some tests delete them)
+bmhuuid1=$(virsh list --all --uuid --name | grep ostest-worker-0 | cut -d ' ' -f 1)
+bmhuuid2=$(virsh list --all --uuid --name | grep ostest-worker-1 | cut -d ' ' -f 1)
+sed "s/CHANGEME/${bmhuuid1}/" ../../common/manifests/bmh-1.yaml | oc apply -n openshift-machine-api -f -
+sed "s/CHANGEME/${bmhuuid2}/" ../../common/manifests/bmh-2.yaml | oc apply -n openshift-machine-api -f -
```
stuggi, line 19: this requires that kuttl runs on the host, or need to configure remote connect for virsh. would it be better to start from an env with 2 bmh configured, expect the VMs are there

```
# oc get bmh -A
NAMESPACE               NAME                 STATE       CONSUMER                      ONLINE   ERROR   AGE
openshift-machine-api   openshift-worker-3   available                                 false            15h
openshift-machine-api   openshift-worker-4   available                                 false            15h
```
Then we take backups of the bmh in a job pre step, which is specific to that osbms testing, so not common prep and that these backups can be used during the tests (delete/create) and als to restore after it finished?

abays: We could do that, but I was assuming we were still being very opinionated on where/how these tests were run.  Basically assuming what we've had here from when we first introduced these KUTTL tests: https://github.com/openstack-k8s-operators/osp-director-operator/blob/548b436a7db2bdfb2dc5452160502815954db622/kuttl-test.yaml#L15

stuggi: right, we require the env be deployed via dev-tools, that is ok but e.g. I usually run the kuttl tests from my local system connected to the host via sshuttle. I can change my workflow, not a problem for ospdo, but e.g. for podified I don't think we can/should do that as kuttl likely has to run on an external node, or pod.

abays: I modified the BaremetalSet KUTTL test to not require the use of `virsh`.  Now the test saves the BMH YAML to `/tmp` and restores it later.

stuggi (approved):
/lgtm

abays:
@stuggi Just had to make one more change to fix something I noticed when running `make openstack` in dev-tools _after_ having run this test

stuggi (approved):
/lgtm

PR #693, opened 2022-11-24
[OSPK8-645] Cleanup host status type

stuggi:
* Splits hoststatus type
    
    Right now osipset and osbms use the HostStatus type, but the osipset
    only use a subset of this. This extracts Hostname, HostRef and
    list of IPAddresses into its own type which then is inline added
    to the HostStatus so there is no change for osbmset. The osipset
    only ise the IPStatus.
    
* Also removes the Ctlplane status as it is no longer used.

* cleanup of kuttl tests. they still work not fully due to https://issues.redhat.com/browse/OSPK8-646


olliewalsh (approved):
/lgtm

PR #703, opened 2022-11-25
[OSPK8-646] fix ip reservation cleanup

stuggi:
*  fix osnetcfg/osnet reservation on role delete
*  cleanup osnetcfg status if there are no reservations in any ctlplane network

    - When single node gets deleted and preserverReservations is set on osnetcfg, the reservations remain in the osnets,  flagged as deleted
    - When the fill role gets deleted and preserverReservations is set on osnetcfg, the reservations remain in the osnets,  flagged as deleted
    - when preserverReservations is or gets flipped to false, the reservations flagged as deleted gets deleted from the osnets

*   [kuttl] update scenarios


olliewalsh (approved):
/lgtm

PR #708, opened 2022-12-01
Use cloud-admin user for local ansible task

olliewalsh:
The username was not set in the inventory for the "undercloud" host so defaulted to root.
The implicit "localhost" in the inventory also defaults to root.

As a result any files copied locally by ansible were owned by root instead of the cloud-admin user. E.g ceph-ansible and stack_action_override.yaml

This sets the correct user for the undecloud host in inventory and switches from localhost to the "undercloud" host for the stack_action_override.yaml generation.


stuggi (approved):
/lgtm

PR #710, opened 2022-12-01
Really use cloud-admin user for local ansible task

olliewalsh:
This appeared to be working when I manually tested it but is not. Turns out the reason root owns the files is tripleo-deploy.sh sets ANSIBLE_BECOME="True". We have to expliciltly disable it if we want the current user.

Fix this for the stack_action_playbook generation. Other cases that hit this such as ceph-ansible will need to be fixed in upstream t-h-t.


stuggi (approved):
/lgtm

PR #719, opened 2022-12-08
update customize_staged_images.sh

stuggi:
Updates customize_staged_images.sh for GA, also only sets the operator related images (osp-director-downloader, osp-director-agent and osp-director-operator) to the midstream registry and leaves other image references in the csv as is.


abays (approved):
/lgtm

dprince (approved):

PR #733, opened 2023-01-05
BLS entries in cloud-init for osbms and osvmset

stuggi:
Per default for the rhel guest image the BLS entries are for machine-id fffffffffffffffffffffffffffffff . When this id does not match the real machine-id from /etc/machine-id updating via grub2-mkconfig won't get set correct.
This updates the entries for /boot/loader/entries/ffffffffffffffffffffffffffffffff-* to match the current machine-id.


stuggi:
alternative or addition to the proposal here https://review.opendev.org/c/openstack/tripleo-ansible/+/855984

PR #739, opened 2023-01-21
Fix ceph export on 17.0

olliewalsh:
The tripleoclient export_ceph logic for cephadm assumes ceph_client.yaml is in specific paths which are not correct for OSPdO. Replicate this logic for now using the correct path for OSPdO until it can be parameterized upstream.

### DIFF
--- a/templates/openstackclient/bin/tripleo-export-ceph
+++ b/templates/openstackclient/bin/tripleo-export-ceph
@@ -1,19 +1,59 @@
 #! /usr/bin/env python3
 
+import logging
 import json
 import os
 import sys
 import tripleoclient.export
 import yaml
 
+logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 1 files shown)


olliewalsh:
17.0 hci subnet job completes successfully with this patch

fultonj (approved):
This looks fine to me. I suppose we could also patch the tripleoclient to better handle this use case though it can get done quickly enough this way.

https://github.com/openstack/python-tripleoclient/blob/master/tripleoclient/export.py#L212

olliewalsh:
> This looks fine to me. I suppose we could also patch the tripleoclient to better handle this use case though it can get done quickly enough this way.
> 
> https://github.com/openstack/python-tripleoclient/blob/master/tripleoclient/export.py#L212

yeah, I will also propose a patch for tripleoclient but doubt I can land it downstream in time for 17.0.1

PR #745, opened 2023-01-25
Update deprecated command and actions in git workflows

olliewalsh:
Node 12 is out of support[1]. Actions using this need to be updated: actions/checkout@v2 -> v3
actions/setup-go@v2 -> v3
golangci/golangci-lint-action@v2 -> v3
tj-actions/branch-names@v5 -> v6

Replace set-output with new env command [2]

Remaining warnings come from redhat-actions which are in the process of being updated.

[1] https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/ [2] https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build-osp-director-operator.yaml
+++ b/.github/workflows/build-osp-director-operator.yaml
@@ -32,7 +32,7 @@ jobs:
       - name: Check secrets are set
         id: have-secrets
         if: "${{ env.imagenamespace != '' }}"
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm

PR #746, opened 2023-01-25
[OSPK8-644] Ignore unassigned hostrefs in VerifyBaremetalStatusHostRefs

olliewalsh:
If an error occurs during the baremetalset reconile loop the status can contain incomplete hostrefs with just the initial state ("unassigned"). The hostref validation in the next reconcile then fails as it expects a BMH with the literal name "unassigned".

This patch updates VerifyBaremetalStatusHostRefs to skip any hostref that are still in this initial state.


abays (commented):
+1 from me, but will let @stuggi take a look first before approving

stuggi (approved):
/lgtm

PR #764, opened 2023-02-17
[OSPK8-666] Ensure bmset finalizer len does not exceed limit

olliewalsh:
The finalizer name must not exceed 63 characters.
Shorten the prefix and use the uid to ensure this is the case.


abays (commented):
+1 from me

abays (approved):
/lgtm

PR #767, opened 2023-02-20
[v1.3.x] [OSPK8-666] Ensure bmset finalizer len does not exceed limit

olliewalsh:
The finalizer name must not exceed 63 characters.
It doesn't have to be globally unique, so just drop the instance name from the finalizer name.


abays (approved):
/lgtm

PR #768, opened 2023-02-22
[OSPK8-672] fix provisionserver localImageUrl update

olliewalsh:
Also set deployment strategy to recreate and stop logging a nil err.

### DIFF
--- a/controllers/openstackprovisionserver_controller.go
+++ b/controllers/openstackprovisionserver_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"net/url"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


controllers/openstackprovisionserver_controller.go
```diff
@@ -269,6 +248,14 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
 				break
 			}
 		}
+
+		// Ensure Status.LocalImageURL is up to date
+		localImageURL := r.getLocalImageURL(instance)
+		if localImageURL != instance.Status.LocalImageURL {
+			// Update status with LocalImageURL
+			common.LogForObject(r, fmt.Sprintf("OpenStackProvisionServer LocalImageURL changed: %s", localImageURL), instance)
+		}
```
abays, line 257: Won't this always fire because L212 will always set `instance.Status.LocalImageURL = ""` whenever we reconcile?

olliewalsh: yeah, had it working then moved it around... will undo

controllers/openstackprovisionserver_controller.go
```diff
@@ -252,12 +231,12 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
 		if podListLen > 1 {
 			common.LogForObject(r, fmt.Sprintf("WARNING: Multiple pods (%d) found for OpenStackProvisionServer %s!", podListLen, instance.Name), instance)
 		} else if podListLen < 1 {
+			// Do not report LocalImageURL while deploying
+			instance.Status.LocalImageURL = ""
```
olliewalsh, line 235: Still not quite right, gets cleared for a seconds but gets set as soon as the pod exists but is not ready to serve

controllers/openstackprovisionserver_controller.go
```diff
@@ -266,6 +244,13 @@ func (r *OpenStackProvisionServerReconciler) Reconcile(ctx context.Context, req
 				cond.Type = shared.ProvisionServerCondTypeProvisioned
 				cond.Reason = shared.OpenStackProvisionServerCondReasonProvisioned
 				cond.Message = "OpenStackProvisionServer has been provisioned"
+
+				// Only set the localImageURL if the pod has finished init
```
olliewalsh, line 248: Working now, localImageURL gets cleared while provisioning (and qemu-img convert is running) and set once the pod is fully initialized

abays (approved):
/lgtm

PR #796, opened 2023-04-17
17.1 support

stuggi:


abays (commented):
+1 from me, but probably want a second pair of eyes to confirm

templates/openstackconfiggenerator/config/17.1
```diff
@@ -0,0 +1 @@
+17.0
```
stuggi, line 1: for now we just link 17.1 -> 17.0 until we see there is a need to have dedicated 17.1 deplate.

olliewalsh (approved):
/lgtm

PR #817, opened 2023-04-26
Fix nic template for split ctlplane

stuggi:
If nodes won't have the External network, like in NovaControl use case, the current nic template won't work. This changes the condition to check External network against role_networks and correct nic index for it.


abays (approved):
/lgtm

olliewalsh (approved):
/lgtm

PR #827, opened 2023-05-17
FFU: Support multiple playbook for a single deploy

olliewalsh:


abays (commented):
+1 from me

stuggi (commented):
Can we do that change since openstackdeploy is an ephemeral resource? what happens if someone applies the same CR from before that change. the `Playbook` will be ignored and we add a playbook based on the mode. could this result in a bad/breaking deployment? wondering if we should keep `Playbook`, flag it as deprecated and have a validation to stop when it is used to mention "use Playbooks instead"?

olliewalsh:
> Can we do that change since openstackdeploy is an ephemeral resource? what happens if someone applies the same CR from before that change. the `Playbook` will be ignored and we add a playbook based on the mode. could this result in a bad/breaking deployment? wondering if we should keep `Playbook`, flag it as deprecated and have a validation to stop when it is used to mention "use Playbooks instead"?

Ah, the unknown field is just dropped instead of rejecting the CR, will look into this

controllers/openstackdeploy_controller.go
```diff
@@ -164,14 +164,22 @@ func (r *OpenStackDeployReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 
 	advancedSettings := instance.Spec.AdvancedSettings.DeepCopy()
-	if advancedSettings.Playbook == "" {
+	if len(advancedSettings.Playbooks) == 0 {
 		switch instance.Spec.Mode {
 		case "update":
-			advancedSettings.Playbook = "update_steps_playbook.yaml"
+			advancedSettings.Playbooks = []string{"update_steps_playbook.yaml"}
+		case "upgrade":
+			advancedSettings.Playbooks = []string{
```
pinikomarov, line 172: isn't  "upgrade-prepare.yaml" supposed to be here ? 

https://github.com/openstack/tripleo-heat-templates/blob/master/environments/lifecycle/upgrade-prepare.yaml#L12

olliewalsh: No, the equivalent is the configgenerator

stuggi (approved):
/lgtm

PR #830, opened 2023-05-24
WIP V1.3.x update to 17.1

yazug:


bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=osp-director-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.18.0+git
```
abays, line 9: Looks like this was somehow downgraded from `1.19` to `1.18`

yazug: yea not sure how or what is doing that ... any thoughts on how to figure out in my system which tools/command is doing this?

PR #844, opened 2023-06-07
Grow volumes on hardened overcloud-full image

olliewalsh:


pkg/common/const.go
```diff
@@ -51,3 +51,37 @@ const (
 	// TripleORolesDataFile -
 	TripleORolesDataFile = "roles_data.yaml"
 )
+
+// GetRoleGrowvolsArgs - return default growvols args for the given tripleo role name
+func GetRoleGrowvolsArgs(role string) []string {
+	if role == "Controller" {
```
stuggi, line 57: should this be a regexp to match all the different default `Controller*` roles?

templates/baremetalset/cloudinit/userdata
```diff
@@ -15,3 +15,6 @@ chpasswd:
 bootcmd:
   # fix BLS entries
   - set -x; if [ -e /boot/loader/entries/ffffffffffffffffffffffffffffffff-* ]; then MACHINEID=$(cat /etc/machine-id) && rename "ffffffffffffffffffffffffffffffff" "$MACHINEID" /boot/loader/entries/ffffffffffffffffffffffffffffffff-* ; fi
+runcmd:
+  # grow volumes on hardened image if command exists
+  - set -x; if [ -e /usr/local/sbin/growvols ]; then /usr/local/sbin/growvols --yes {{ range .GrowvolsArgs}}{{.|trim}} {{end}}; fi
```
stuggi, line 20: from where is  `/usr/local/sbin/growvols` coming? is it part of the overcloud image?

olliewalsh: yes

olliewalsh: only the overcloud images with lvm. Logic comes from  https://github.com/openstack/tripleo-ansible/blob/stable/wallaby/tripleo_ansible/playbooks/cli-overcloud-node-growvols.yaml which is run on all nodes when deploying with python-tripleoclient

pkg/common/const.go
olliewalsh: this is copied from the playbook used in a traditional director deployment so I've kept is consistent

steveb (commented):
/lgtm

Even if the underlying mechanism switches from cloud-init to ansible in the future, the exposed GrowvolsArgs interface won't change, so this is fine.

The only downside of cloud-init is if they provide custom invalid GrowvolsArgs, and they may not notice until the deployment fails with a full (tiny) disk. I don't think this will be a huge issue in practice

PR #845, opened 2023-06-12
Switch OsProvServer perms to 'hostnetwork' instead of 'privileged'

abays:
Takes inspiration from https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/22 to improve security of the provision server pod by downgrading it from `privileged` to only `hostnetwork`.  Also further improves security by removing the `privileged` and `anyuid` permissions from the controller-manager pod, as they are not needed.

### DIFF
--- a/controllers/openstackprovisionserver_controller.go
+++ b/controllers/openstackprovisionserver_controller.go
@@ -91,8 +91,6 @@ func (r *OpenStackProvisionServerReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;update;watch;
 // +kubebuilder:rbac:groups=machine.openshift.io,resources="*",verbs="*"
 // +kubebuilder:rbac:groups=metal3.io,resources="*",verbs="*"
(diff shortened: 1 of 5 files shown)


controllers/openstackprovisionserver_controller.go
```diff
@@ -91,8 +91,6 @@ func (r *OpenStackProvisionServerReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;update;watch;
 // +kubebuilder:rbac:groups=machine.openshift.io,resources="*",verbs="*"
 // +kubebuilder:rbac:groups=metal3.io,resources="*",verbs="*"
-// +kubebuilder:rbac:groups=security.openshift.io,namespace=openstack,resources="securitycontextconstraints",resourceNames="privileged",verbs="use"
-// +kubebuilder:rbac:groups=security.openshift.io,namespace=openstack,resources="securitycontextconstraints",resourceNames="anyuid",verbs="use"
```
stuggi, line 95: do we need ?
```
// +kubebuilder:rbac:groups="security.openshift.io",resourceNames=hostnetwork,resources=securitycontextconstraints,verbs=use
```

PR #864, opened 2023-06-28
[OSPK8-439] Stop trying to parse git url, just set GIT_SSH_COMMAND

olliewalsh:
The only reason the git url is parsed is to generate an ssh config file.
Since parsing git urls is non-trivial we can avoid this completely and
just set the required SSH config on the command line via GIT_SSH_COMMAND.

Also remove the hardcoded git user from GIT_SSH_COMMAND

### DIFF
--- a/templates/openstackconfiggenerator/bin/create-playbooks.sh
+++ b/templates/openstackconfiggenerator/bin/create-playbooks.sh
@@ -7,16 +7,7 @@ cp /mnt/ssh-config/* $HOME/.ssh/
 chmod 600 $HOME/.ssh/git_id_rsa
 chown -R cloud-admin: $HOME/.ssh
 
-GIT_HOST=$(echo $GIT_URL | sed -e 's|^git@\(.*\):.*|\1|g')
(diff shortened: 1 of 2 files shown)


templates/openstackconfiggenerator/bin/create-playbooks.sh
```diff
@@ -7,16 +7,7 @@ cp /mnt/ssh-config/* $HOME/.ssh/
...
-GIT_USER=$(echo $GIT_URL | sed -e 's|^git@.*:\(.*\)/.*|\1|g')
-
-cat <<EOF > $HOME/.ssh/config
-Host $GIT_HOST
-    User $GIT_USER
-    IdentityFile $HOME/.ssh/git_id_rsa
-    StrictHostKeyChecking no
-EOF
-chmod 644 $HOME/.ssh/config
+export GIT_SSH_COMMAND="ssh -i $HOME/.ssh/git_id_rsa -l git -o StrictHostKeyChecking=no"
```
stuggi, line 10: is it ok to have fixed user `-l git` ?

stuggi (approved):
/lgtm

PR #880, opened 2023-07-25
[OSPK8-735] Improve logging for OSBMS scale-up

abays:
Be more explicit about what is happening during a scale-up of an `OpenStackBaremetalSet`

### DIFF
--- a/api/v1beta1/common_openstackbaremetalset.go
+++ b/api/v1beta1/common_openstackbaremetalset.go
@@ -102,7 +102,7 @@ func VerifyBaremetalStatusHostRefs(
 			}
 
 			if !found {
-				err := fmt.Errorf("Existing BaremetalHost \"%s\" not found for OsBaremetalSet %s.  "+
+				err := fmt.Errorf("Existing BaremetalHost \"%s\" not found for OpenStackBaremetalSet %s.  "+
 					"Please check BaremetalHost resources and re-add \"%s\" to continue",
 					bmhStatus.HostRef, instance.Name, bmhStatus.HostRef)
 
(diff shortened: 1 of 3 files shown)


api/v1beta1/common_openstackbaremetalset.go
```diff
@@ -144,12 +162,15 @@ func VerifyBaremetalSetScaleUp(log logr.Logger, instance *OpenStackBaremetalSet,
 
 		// If we can't satisfy the new requested replica count, explicitly state so
 		if newBmhsNeededCount > len(availableBaremetalHosts) {
-			return nil, fmt.Errorf("Unable to find %d requested BaremetalHost count (%d in use, %d available)",
+			return nil, fmt.Errorf("Unable to find %d requested BaremetalHost count (%d in use, %d available)%s for OsBaremetalSet %s",
```
olliewalsh, line 165: I would use OpenstackBaremetalSet vs OsBaremetalSet to avoid confusion (or just use the Kind here)

olliewalsh (changes requested):

api/v1beta1/common_openstackbaremetalset.go
abays: I was maintaining parity with other places where I had apparently used `OsBaremetalSet` (such as L105), but yeah, it would be better to just spell it out or use Kind.

api/v1beta1/common_openstackbaremetalset.go
```diff
@@ -123,17 +123,35 @@ func VerifyBaremetalSetScaleUp(log logr.Logger, instance *OpenStackBaremetalSet,
...
+		log.Info(fmt.Sprintf("Attempting to find %d BaremetalHost(s)%s for scale-up of OpenStackBaremetalSet %s", newBmhsNeededCount, labelStr, instance.Name))
+
 		for _, baremetalHost := range allBmhs.Items {
-			if baremetalHost.Spec.Online || baremetalHost.Spec.ConsumerRef != nil {
-				continue
+			mismatch := false
+
+			if baremetalHost.Spec.Online {
+				log.Info(fmt.Sprintf("BaremetalHost %s cannot be used because it is already online", baremetalHost.ObjectMeta.Name))
+				mismatch = true
```
stuggi, line 140: does it make sense to continue here? I guess if a BMH is online it has a consumerRef and we do not have to check for hardware req, or can there be a situation where this is not the case?

api/v1beta1/common_openstackbaremetalset.go
```diff
@@ -123,17 +123,35 @@ func VerifyBaremetalSetScaleUp(log logr.Logger, instance *OpenStackBaremetalSet,
...
+			}
+
+			if baremetalHost.Spec.ConsumerRef != nil {
+				log.Info(fmt.Sprintf("BaremetalHost %s cannot be used because it already has a consumerRef", baremetalHost.ObjectMeta.Name))
+				mismatch = true
```
stuggi, line 145: wondering if it is ok to just continue instead of set mismatch?

api/v1beta1/common_openstackbaremetalset.go
abays: The reason I am not just continuing is that the customer asked specifically for a detailed list of which things were preventing the BMH from being selected.  So my aim was to gather everything that applied as opposed to just logging the first one.

olliewalsh (approved):

api/v1beta1/common_openstackbaremetalset.go
olliewalsh: hrmm, yes think continuing here might be better. Otherwise would verifyBaremetalSetHardwareMatch would result in a lot of log spam for hosts that are already in use?

abays: I'll revisit this and try to find some balance.  I guess the question is what constitutes "more detail" and what veers into the category of "spam" as far as the RFE is concerned.

PR #887, opened 2023-08-02
Ephemeral heat communication use svc fqdn

stuggi:


controllers/openstackconfiggenerator_controller.go
```diff
@@ -299,7 +299,7 @@ func (r *OpenStackConfigGeneratorReconciler) Reconcile(ctx context.Context, req
 	}
 
 	templateParameters["TripleoDeployFiles"] = tripleoDeployFiles
-	templateParameters["HeatServiceName"] = "heat-" + instance.Name
+	templateParameters["HeatServiceName"] = "heat-" + instance.Name + "." + instance.Namespace + ".svc"
```
olliewalsh, line 302: would it be better to use the entire fqdn with `.svc.cluster.local`?

olliewalsh: I expect then just adding `.cluster_local` to no_proxy would work

PR #899, opened 2023-08-30
Stop setting FSGroup

olliewalsh:
FSGroup recursively chowns/chgrps each volume and setgids the dirs. This can interfere with permissions on the openstackclient, which needs to resemble a traditional undercloud host as closely as possible. SSH in particular has strict rule for the ownership and mode of it's files, but there could be other pemissions issues too.

Note FSGroup has no effect on NFS, which we happen to be using in CI, but will affect local volumes, RBD etc... which provide formatted block devices.

Avoid this completely by removing FSGroup and explicitly setting all ownership/modes in the init container.

Jira: [OSP-27993](https://issues.redhat.com//browse/OSP-27993)
Resolves: rhbz#2236085


stuggi (approved):
/lgtm

PR #900, opened 2023-08-31
Add FFU multi-rhel support

olliewalsh:
Adds tripleoRoleOverride param to config generator to allow a different t-h-t role to be applied to a subset of hosts in heat.
Heat indices are shuffled around to compensate but the actual hostnames remain unchanged.

For example:
```
    tripleoRoleOverride:
      Compute:
        roleName: ComputeRhel8
        hostIndexMap:
          1: true
          3: true
      ComputeFoo:
        roleName: ComputeFooRhel8
        hostIndexMap:
          0: true
          1: false
          2: false
          3: false
          4: false
          5: true
```

compute-0 remains the same in heat
compute-1 becomes computerhel8-0 in heat (still compute-1 in hostnamemap)
compute-2 becomes compute-1 in heat (still compute-2 in hostnamemap)
compute-3 becomes computerhel8-1 (still compute-3 in hostnamemap)
compute-4 becomes compute-2 in heat (still compute-4 in hostnamemap)

Then, as per a regular director-driven deployment, the role params for the new role can be used to override the contain images etc... just for that role to use the rhel8 config.

hostIndexMap entries default to false if not set, but can be set explicitly e.g to keep track of which hosts have already been migrated to rhel9


pkg/openstackconfiggenerator/configmap.go
```diff
@@ -383,7 +384,13 @@ func createRolesMap(
 
 	for _, osnet := range osNetList.Items {
 		for roleName, roleReservation := range osnet.Spec.RoleReservations {
-			if IsRoleIncluded(roleName, instance) {
+			var roleOveride ospdirectorv1beta1.TripleoRoleOverrideSpec
+			roleOverrideFound := false
+			if instance.Spec.TripleoRoleOverride != nil {
```
stuggi, line 389: should we add a check that this only works on OSP 17.1, or this also supported in 16.2?

api/v1beta1/openstackconfiggenerator_types.go
```diff
@@ -50,6 +58,8 @@ type OpenStackConfigGeneratorSpec struct {
 	// - Production OSP environments MUST have fencing enabled.
 	// - Requires the fence-agents-kubevirt package to be installed in the virtual machines for the roles running pacemaker.
 	EnableFencing bool `json:"enableFencing"`
+	// TripleoRoleOverride - map of TripleO role name to temporary role override to support a multi-rhel environment
```
stuggi, line 61: have to mention that this is only supported in 17.1 and later?

pkg/openstackconfiggenerator/configmap.go
olliewalsh: hrmm, well you could use it with 16.2 I guess, but not sure why you would want to

stuggi: so then, to prevent someone using it with 16.2 should we prevent that since the code if for both versions?

olliewalsh: Ok. Validating in a webhook would be very heavy since it would need to get the openstackcontrolplane to check the version. Could just return an error in the reconcile loop if it's not 17.1.

stuggi: yes, that works or we could do the same in the configgenerator webhook, or you say thats the heavy way?
```
if instance.Spec.TripleoRoleOverride != nil && ospVersion != shared.OSPVersion(shared.TemplateVersion17_1) {
		return &TripleoRoleOverrideVersionError{ospVersion}
	}
```

stuggi (approved):
/lgtm

PR #952, opened 2023-11-28
Fix override of rendered files via tarball

stuggi:
Extracting/placing default tht files via the script to render the playbooks is intended to allow to override default and rendered tht files. Right now the ordering to copy the files in the final place first copies the tarball and afterwards the default/rendered ones. This prevents from override them via the tarball.

This changes the order to extract the tarball last.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2251800


abays (approved):
/lgtm

PR #974, opened 2024-01-09
[OSPK8-784] Allow user-provided cloud-init secrets for BMHs

abays:
Introduce the ability of users to pre-configure `BaremetalHost`s with custom cloud-init `Secret`s for user data and/or network data.  If a BMH chosen for `OpenStackBaremetalSet` spin-up/scale-up already has a user data or network data `Secret`, we will not auto-generate it as we normally would and will instead simply apply our "must-gather" label for potential data collection.

I've also improved the efficiency of how we pass `BaremetalHost` objects through the reconciliation logic in the `OpenStackBaremetalSet` controller.  The old form was getting a list of the BMHs and then re-acquiring the individual BMHs that it already had in the list, which was an unnecessary re-read.


controllers/openstackbaremetalset_controller.go
```diff
@@ -788,7 +788,11 @@ func (r *OpenStackBaremetalSetReconciler) ensureBaremetalHosts(
 	newBmhsNeededCount := instance.Spec.Count - len(existingBaremetalHosts.Items)
 
 	// Sort the list of available BaremetalHosts
-	sort.Strings(availableBaremetalHosts)
+	sort.Slice(availableBaremetalHosts[:], func(i, j int) bool {
+		// strings.Compare(a, b) returns -1 when a < b, which is the comparison we want
+		// for ordering alphabetically as sort.Slice expects a "less" function
+		return strings.Compare(availableBaremetalHosts[i].Name, availableBaremetalHosts[j].Name) == -1
```
olliewalsh, line 794: Expect you could just use <, that what's used for StringSlice https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/sort/sort.go;l=154

abays: Ack, I will try that instead to keep things simpler

pkg/baremetalset/const.go
```diff
@@ -20,9 +20,6 @@ const (
 	// AppLabel -
 	AppLabel = "osp-baremetalset"
 
-	// BaremetalHostRemovalAnnotation - Annotation key placed on BaremetalHost resources to target them for OpenStackBaremetalSet scale-down
-	BaremetalHostRemovalAnnotation = "osp-director.openstack.org/baremetalset-delete-baremetalhost"
-
```
abays, line 25: This is old and not used.  We use this [1] instead.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/b5d119de2c7998c03061cdf1d8f5ea1ccf9d4849/api/shared/annotations.go#L4-L5

olliewalsh (approved):
/lgtm

PR #981, opened 2024-01-15
Fix osconfigenerator for Redis/OVNDBs vips

stuggi:
Special Redis on OVNDBs roles are not considered in the configgenerator when rendering the template. As a result the required hiera redis_vip was not set

Function lookup() did not find a value for the name 'redis_vip'

This updates the configgenerator to consider Redis and OVNDBs and have the entries like the following in the rendered parameter file:

~~~
  OVNDBsVirtualFixedIPs:
    - ip_address: 172.17.0.12
      use_neutron: false
  RedisVirtualFixedIPs:
    - ip_address: 172.17.0.11
      use_neutron: false
 ~~~

### DIFF
--- a/pkg/openstackconfiggenerator/configmap.go
+++ b/pkg/openstackconfiggenerator/configmap.go
@@ -357,9 +357,16 @@ func createNetworksMap(
 }
 
 // IsRoleIncluded - checks if the role exists in the ConfigGenerator Roles set
-func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator) bool {
+func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator, ospVersion shared.OSPVersion) bool {
 
-	if len(instance.Spec.Roles) == 0 || roleName == controlplane.Role {
+	// return for special conditions:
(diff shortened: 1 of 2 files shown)


pkg/openstackconfiggenerator/configmap.go
```diff
@@ -357,9 +357,10 @@ func createNetworksMap(
 }
 
 // IsRoleIncluded - checks if the role exists in the ConfigGenerator Roles set
-func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator) bool {
+func IsRoleIncluded(roleName string, instance *ospdirectorv1beta1.OpenStackConfigGenerator, ospVersion shared.OSPVersion) bool {
 
-	if len(instance.Spec.Roles) == 0 || roleName == controlplane.Role {
+	if len(instance.Spec.Roles) == 0 || roleName == controlplane.Role ||
+		(ospVersion == shared.OSPVersion(shared.TemplateVersion17_1) && (roleName == "Redis" || roleName == "OVNDBs")) {
```
olliewalsh, line 363: maybe just add a comment to explain

olliewalsh (approved):
/lgtm

PR #994, opened 2024-02-01
bump manager default limits

stuggi:


stuggi:
OOMKilled seem in environments, lets bump it

config/manager/manager.yaml
```diff
@@ -34,8 +34,8 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 768Mi
+            memory: 1Gi
```
olliewalsh, line 37: should we make it higher?

stuggi: changed to 2Gi

olliewalsh: Wonder if we could test this to get a rough idea of the memory requirements? go-git has open issues for using lots of memory....

olliewalsh (approved):
/lgtm

PR #1013, opened 2024-03-15
[v1.3.x] add .cluster.local to ephemeral heat urls

stuggi:


abays (approved):
Tested in a couple environments and it works

/lgtm

PR #1014, opened 2024-03-18
Use bundle digest in build script for offline compatibility

abays:
Without this, trying to mirror builds for offline use fails like so:

```
$ oc adm catalog mirror ${INDEX_IMG} ${LOCAL_REGISTRY} --insecure --index-filter-by-os='Linux/x86_64'
...
no digest mapping available for quay.io/andrewbays/osp-director-operator-index:1.3.0-16.2, skip writing to ImageContentSourcePolicy
no digest mapping available for quay.io/andrewbays/osp-director-operator-bundle:1.3.0-16.2, skip writing to ImageContentSourcePolicy
```

Also changes the agent image name from `osp-director-operator-agent` to `osp-director-agent` because the latter is used everywhere else.

### DIFF
--- a/scripts/build_and_push_images.sh
+++ b/scripts/build_and_push_images.sh
@@ -20,7 +20,7 @@ IMG="$REPO/$OP_NAME":$VERSION
 IMAGE_TAG_BASE="$REPO/$OP_NAME"
 
 
-AGENT_IMAGE="$OP_NAME-agent"
+AGENT_IMAGE="osp-director-agent"
 AGENT_IMG_BASE="$REPO/$AGENT_IMAGE"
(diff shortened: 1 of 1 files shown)


scripts/build_and_push_images.sh
```diff
@@ -20,7 +20,7 @@ IMG="$REPO/$OP_NAME":$VERSION
 IMAGE_TAG_BASE="$REPO/$OP_NAME"
 
 
-AGENT_IMAGE="$OP_NAME-agent"
+AGENT_IMAGE="osp-director-agent"
```
abays, line 23: This is the format used everywhere else as opposed to `osp-director-operator-agent`

dprince (approved):

PR #1043, opened 2024-05-02
[OSPK8-803] Use list format for SSH authorized keys in OSBMS cloud-init

abays:
The proper format for authorized keys in cloud-init is a [list](https://cloudinit.readthedocs.io/en/latest/reference/examples.html#configure-instance-s-ssh-keys), but we are using a string instead.  As a result, if multiple keys are desired, none of the SSH keys will work if we place them in that string.


stuggi (approved):
/lgtm

PR #1056, opened 2024-08-08
Bump golang to 1.21

stuggi:
* bumps golang to 1.21
* removes deprecated math_rand.Seed() call
* bumps image_downloader to use centos:stream9
* minor updates to Makefile and bump kube-rbac-proxy image

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
         args:
         - "--secure-listen-address=0.0.0.0:8443"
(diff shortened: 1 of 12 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1
```
dprince, line 13: this URL will go away soon I think. Fine to do this now as it matches what we do in next gen. But soon we should revisit these and either use the downstream image for kube-rbac-proxy or perhaps look at options to remove it

dprince: to be clear, no action needed in this PR at this time

stuggi:
job using quay.io/mschuppe/osp-director-operator:bump_golang-latest-17.1 passed fine

abays (approved):
/lgtm

PR #1096, opened 2024-10-22
Trust openstackcontrolplane CAs in configgenerator

olliewalsh:
Add the openstackcontrolplane CAConfigMap CA certs to the trusted certs when running the config generator job as it may need to connect to container registries that use a custom CA (e.g to resolve image tags).


stuggi (approved):
/lgtm

PR #1098, opened 2024-10-24
Return error id OpenStackNet misses OpenStackNetConfig owner ref

stuggi:
In case of backup restore the owner ref needs to get updated by the OpenStackNetConfig controller. If some other reconcile using the OpenStackNet happens first, this reference is missing and e.g. the OpenStackIPSet gets stuck waiting for IPs to be processed.


abays (approved):
Tested with greenfield and backup/restore.  Succeeded.

/lgtm

PR #1100, opened 2024-10-28
Make IPSet OSclient owner check more robust

stuggi:
If the owner of an OSIPset is an OpenStackClient resource, the role name of an role reservation gets set to
<openstackclient.Role><instance.Name>. Right now this check is using an owner label, and queries if an OpenStackClient object exist with this name. In some circumstance this has seen to fail, most likely on the initial create and as a result it ended in two role reservations in the OSNet.

~~~
  roleReservations:
    OpenstackClientopenstackclient: <<<<<<<<<<<< good reservation
      addToPredictableIPs: false
      reservations:
      - deleted: false
        hostname: openstackclient-0
        ip: x.x.x.x
        serviceVIP: false
        vip: false
...
    openstackclient:  <<<<<<<<<<<< wrong reservation
      addToPredictableIPs: false
      reservations:
      - deleted: true
        hostname: openstackclient-0
        ip: x.x.x.x
        serviceVIP: false
        vip: false
status:
  reservations:
  ...
    openstackclient-0:
      deleted: true <<<<<<<<<<<<<<<< wrong marked as deleted openstackclient above
      ip: x.x.x.x
~~~

Because later reconciliation have the correct role, the later one is marged as deleted and as a result the reservation in the status get set to delete: true, too. This results in the corresponding IPSet to never reach the success state.

This makes the check if the owner is an OpenStackClient more robust, by checking the owner ref lists controller entry and compares if it is `OpenStackClient`. With this there is no api call required, which could fail for different reasons.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=2321103

### DIFF
--- a/controllers/openstacknetconfig_controller.go
+++ b/controllers/openstacknetconfig_controller.go
@@ -1228,22 +1228,11 @@ func (r *OpenStackNetConfigReconciler) ensureIPReservation(
 		//
 		// For backward compatability check if owning object is osClient and set Role to <openstackclient.Role><instance.Name>
 		//
-		osClient := &ospdirectorv1beta1.OpenStackClient{}
-		err := r.Get(ctx, types.NamespacedName{
-			Name:      osIPset.Labels[common.OwnerNameLabelSelector],
-			Namespace: osIPset.Namespace},
-			osClient)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


abays (commented):
Seems fine to me at first glance

dprince (approved):

PR #1102, opened 2024-10-29
Improve backup and restore functionality for OpenStackBaremetalSets

abays:
When we reworked the way that we handle `OpenStackBaremetalSet` scale-up a while back, we inadvertently introduced logic that could prevent `OpenStackBackup`s from being restored.  We need to make sure that `cleanRestore` backups first allow any existing in-use BMHs to deprovision during cleaning before transitioning to the actual restore.  Failure to do so will possibly cause our newer `OpenStackBaremetalSet` scale-up logic to fail if there are not enough BMHs in the state `available` state to satisfy all `OpenStackBaremetalSet` requests.  If that logic fails, the restore process will error-out and be stuck.

### DIFF
--- a/pkg/openstackbackup/funcs.go
+++ b/pkg/openstackbackup/funcs.go
@@ -8,6 +8,8 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 
+	metal3v1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/openstack-k8s-operators/osp-director-operator/api/shared"
 	ospdirectorv1beta1 "github.com/openstack-k8s-operators/osp-director-operator/api/v1beta1"
 	ospdirectorv1beta2 "github.com/openstack-k8s-operators/osp-director-operator/api/v1beta2"
 	"github.com/openstack-k8s-operators/osp-director-operator/pkg/common"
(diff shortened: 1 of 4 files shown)


pkg/openstackbackup/funcs.go
```diff
@@ -605,11 +607,94 @@ func CleanNamespace(
...
+					}
+				}
+			}
+		}
+
+		// If we added our "RequireBMHDeprovision" label to any BMH, we need to return from the reconcile
+		// and then check back again later
+		if waitingOnBmhDeprovisionLabelAdded {
+			return false, nil
+		}
```
abays, line 653: I'm not sure this is really necessary.  If the label was added to one of more BMHs, then it should be okay to proceed to the `DeleteAll()` of the `OpenStackBaremetalSet`s below.

pkg/openstackbackup/funcs.go
```diff
@@ -605,11 +607,94 @@ func CleanNamespace(
...
+						return false, err
+					}
+				}
+			}
+		}
+
+		// If we added our "RequireBMHDeprovision" label to any BMH, we need to return from the reconcile
+		// and then check back again later
+		if waitingOnBmhDeprovisionLabelAdded {
+			return false, nil
```
stuggi, line 652: do we need to return an requeueAfter since the backup controller is not watching the BMHs? I mean could there be a possibility that there is no reconcile event and we are just waiting for one to happen?

pkg/openstackbackup/funcs.go
```diff
@@ -605,11 +607,94 @@ func CleanNamespace(
...
+	)
+	if err != nil {
+		return false, err
+	}
+
+	for _, bmh := range baremetalHostsList.Items {
+		// If the BMH is not available (i.e. not done deprovisioning), we need to return from
+		// this reconcile and check back again later
+		if bmh.Status.Provisioning.State != metal3v1.StateAvailable {
+			return false, nil
```
stuggi, line 680: same as above

pkg/openstackbackup/funcs.go
stuggi: cool

stuggi:
change sounds good to me

dprince (approved):

PR #1117, opened 2024-12-05
Fix backup/restore KUTTL test to work with current OSBMS config sample

abays:
We changed the `OpenStackBaremetalSet` `samples/config` YAML a while ago and need to adjust the backup/restore KUTTL test so that it works with it properly


stuggi (approved):
/lgtm

PR #1118, opened 2024-12-06
fix AddOSNetNameLowerLabels for network delete

stuggi:
also adds functional test for AddOSNetNameLowerLabels and logs for IPset label update


controllers/openstackipset_controller.go
```diff
@@ -166,9 +166,6 @@ func (r *OpenStackIPSetReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		instance.Labels,
 		instance.Spec.Networks,
 	)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
```
stuggi, line 171: `AddOSNetNameLowerLabels` does not return an err

abays (approved):
/lgtm

PR #1131, opened 2025-01-08
[OSPRH-12353] Prevent nil pointer panic when network validation fails

abays:
If an `OpenStackNet` is missing, we don't have an actual object to reference for producing an error message.  Therefore, just hard-code the Kind into the message, since it won't ever be anything other than `OpenStackNet`.


stuggi (approved):
/lgtm

PR #1143, opened 2025-02-14
[OSPRH-14001]Make fields optional in v1beta1 to resolve reinstall issues

stuggi:
With 4.16 there is additional validation in olm when the installplan gets deployed. In this validation the CR gets checked against the old v1beta1 CRD, even the object storage version is v1beta2. As there are no conversion webhooks in ospdo the verification fails.

To resolve this this adds the fields which get reported as errors in v1beta1 as optional.

### DIFF
--- a/api/v1beta1/openstackcontrolplane_types.go
+++ b/api/v1beta1/openstackcontrolplane_types.go
@@ -38,13 +38,13 @@ type OpenStackControlPlaneSpec struct {
 	// OpenStackClientNetworks the name(s) of the OpenStackClientNetworks used to attach the openstackclient to
 	OpenStackClientNetworks []string `json:"openStackClientNetworks"`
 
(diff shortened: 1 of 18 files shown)


api/v1beta1/openstackcontrolplane_types.go
```diff
@@ -38,13 +38,13 @@ type OpenStackControlPlaneSpec struct {
 	// OpenStackClientNetworks the name(s) of the OpenStackClientNetworks used to attach the openstackclient to
 	OpenStackClientNetworks []string `json:"openStackClientNetworks"`
 
-	// +kubebuilder:default=false
```
olliewalsh, line 41: bool defaults to false anyway

olliewalsh (approved):
/lgtm

PR #1147, opened 2025-02-17
[vmset] allow setting eviction and run strategy

stuggi:


config/crd/bases/osp-director.openstack.org_openstackbackups.yaml
```diff
@@ -3983,6 +3983,15 @@ spec:
                                           GB - use RootDisk.DiskSize instead
                                         format: int32
                                         type: integer
+                                      evictionStrategy:
+                                        default: LiveMigrate
+                                        description: VirtualMachineRunStrategy is
```
olliewalsh, line 3988: maybe we should override the descriptions?

controllers/openstackvmset_controller.go
```diff
@@ -783,7 +783,7 @@ func (r *OpenStackVMSetReconciler) vmCreateInstance(
 	osNetBindings map[string]ospdirectorv1beta1.AttachType,
 ) error {
 
-	evictionStrategy := virtv1.EvictionStrategyLiveMigrate
+	evictionStrategy := virtv1.EvictionStrategy(instance.Spec.EvictionStrategy)
```
olliewalsh, line 786: could default to LiveMigrate only when storage is RWX

config/crd/bases/osp-director.openstack.org_openstackbackups.yaml
stuggi: updated

controllers/openstackvmset_controller.go
stuggi: updated

controllers/openstackvmset_controller.go
```diff
@@ -1021,6 +1020,16 @@ func (r *OpenStackVMSetReconciler) vmCreateInstance(
 			instance.Namespace,
 		)
 
+		if instance.Spec.EvictionStrategy != nil {
+			vm.Spec.Template.Spec.EvictionStrategy = instance.Spec.EvictionStrategy
+		} else {
+			vm.Spec.Template.Spec.EvictionStrategy = ptr.To(virtv1.EvictionStrategyLiveMigrate)
+		}
+		if instance.Spec.RootDisk.StorageAccessMode == "ReadWriteOnce" {
```
olliewalsh, line 1028: only when instance.Spec.EvictionStrategy == nil

olliewalsh (approved):
/lgtm

PR #1161, opened 2025-03-13
bump golang to 1.23, golang.org/x/crypto to v0.36.0 and github.com/go-git/go-git/v5 to v5.14.0

stuggi:


stuggi:
we'd have to pump golangci-lint to v1.63.4 to support golang 1.23. otherwise it will fail with the OOM

stuggi:
golangci issue should be resolved with https://github.com/openshift/release/pull/62784

PR #1259, opened 2026-03-16
Pin golang.org/x/crypto and gophercloud/v2 to avoid Go 1.25 requirement

stuggi:
golang.org/x/crypto >= 0.49.0 and github.com/gophercloud/gophercloud/v2
>= 2.11.0 require Go 1.25, which is not yet available. Pin these
dependencies in dependabot to prevent incompatible upgrades.


abays (approved):
/lgtm

PR #1311, opened 2026-08-04
Enable TLS verification in fencing kubeconfig

abays:
The fencing kubeconfig template hardcoded `insecure-skip-tls-verify: true`, allowing a network-adjacent attacker to impersonate the API server and capture the kubevirt fencing agent's `ServiceAccount` token. Extract the cluster CA from the operator's `rest.Config` and template it as `certificate-authority-data` so the fencing agent validates the API server certificate.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/controllers/openstackvmset_controller.go
+++ b/controllers/openstackvmset_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


abays:
Successfully tested locally via a new operator build, subsequent OOO deployment and `pcs stonith fence` command.

abays:
This will break fencing functionality if/when the cluster rotates its certs, as there is no automated mechanism by which to provide the new cert to the `Secret`.

PR #1312, opened 2026-08-05
Narrow wildcard RBAC to least-privilege

abays:
Remove `machine.openshift.io` and `hco.kubevirt.io` wildcard RBAC markers that granted full access to API groups with no corresponding code usage. Replace the `metal3.io` wildcard with a scoped rule granting only get on `provisionings`, which is the single resource actually accessed via the dynamic client.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -124,13 +124,6 @@ rules:
   - list
   - watch
 - apiGroups:
-  - machine.openshift.io
-  - metal3.io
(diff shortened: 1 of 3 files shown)


abays:
Successfully tested locally via a new operator build, subsequent OOO deployment.

stuggi (approved):
/lgtm

